PR #13, opened 2017-09-08
Clean up assert handling

bigpoint-qlansu:
The main goal of this pull request is to clean up the assert handling so it be more easily extended in additional pull requests.

I've changed the parameters of the assert handling bus to a struct reference, so parameters can be added or removed later without affecting hundreds of files. The methods now also use a return code instead of a boolean, which allows the assert handler to store ignored asserts in a single location, across all DLLs.

Finally, this change allows the debugger to break on the line of the assert in the affected code instead of inside the assert handling method.


AMZN-alexpete:
Hi @bigpoint-qlansu,
Thank you for submitting this fix!  We have created and assigned the issue to a developer.  The internal issue for tracking purposes is LY-63340.

AMZN-alexpete:
Hi @bigpoint-qlansu! Just wanted to let you know although your pull request has been closed, we are reviewing it and will get back to you once we've determined whether we can accept it. Thank you kindly for your contribution!

PR #18, opened 2017-09-15
Fixed asset name display

alex-leleka:
convert assetFilename to const char * for error log function.


ranilianaws:
Hello @alex-leleka,
Thanks for submitting! Although we have closed the request, we are currently reviewing it and will get back to you once we've determined we can accept it. Thanks for contributing to Lumberyard!

ranilianaws:
Hello @alex-leleka ,
We've modified the code to the following as that seems to be the encoding we are going forward with:
`assetFilename.toUtf8().data()`

Thank you again for your contribution.

PR #21, opened 2017-09-22
bugfix. Unable to save View Dist Ratio for "Designer Objects" (Box, Cilinder, etc...). Property was serialized to "ViewDistMultiplier" attribute, but read from "ViewDistRatio"

LatzekLat:


sourjon:
Hi @LatzekLat,

Thanks for your submission!  This particular component is currently considered legacy, so we won't be updating it any further.  That said, we really appreciate your contribution, and thanks for helping improve Lumberyard!

PR #22, opened 2017-09-29
FillSpaceTool creates backfacing faces (Render camera is not valid)

LatzekLat:
1) Create designer object (Cube)
2) Delete few faces(one that is looking into positive X direction, another that is looking into opposite)
3) Recreate both faces. Face that is looking to X negative direction will face opposite from camera direction, and will be clipped.


sourjon:
Hi @LatzekLat ,

Thanks for your contribution to Lumberyard!  Because this system is part of the Legacy Rollup Bar, we won't be taking this change for a future release, but we do appreciate your effort, and we'll look for more of your updates in the future!

PR #25, opened 2017-11-06
Command 'r_measureOverdraw' fix

alex-leleka:
r_measureOverdraw doesn't work if active render target is backbuffer, so we switch to back buffer at the very end after all debug rendering has been done.


ranilianaws:
Hello @alex-leleka,
We reviewed the code and proposed a different approach - instead of changing the order of operations we did the following:

- Removed referrences to pRT in DriverD3D.cpp
- Added CTexture::s_ptexSceneDiffuse->Apply(0, CTexture::GetTexState(TexStatePoint)); after FX_Commit().

That seemed to also fix the r_measureOverdraw issue :)

PR #26, opened 2017-11-14
Change destroy loop bounds in CryArray

alex-leleka:
Iterator end() point to an element after the last element in array, so calling destructor for it isn't secure.


ranilianaws:
Hello @alex-leleka,
We are currently reviewing the code and will get back to you as soon as possible.

ranilianaws:
Hello @alex-leleka.

We have accepted the code into the codebase. It will be available in a future version of Lumberyard.

Thank you again for your contribution!

PR #28, opened 2017-11-28
Support build version for linux build

Ximura:
Using --force-version in lmbrwaf build doesn't change version for linux build

**lmbrwaf changes:**

CryEngine use defines EXE_VERSION_INFO_# to propagate build version in linux build, for this lmbrwaf should add defines to CrySystem with value from self.bld.get_game_project_version() (in case --force-version doesn't used it would be version of lumberyard)
Currently lmbrwaf pass EXE_VERSION_INFO_# defines with game project version only to launcher, but doesn't used in launcher project and linux build in result have lumberyard version, but it's expected to get version provided to --force-version 
As we don't need to add this define to every project, I decided to add new feature that would be used only in CrySystem, so that we don't recompile whole project in case version changed.

**SystemCFG.cpp changes:**
CryEngine build system patched CrySystem with build version on Windows, that why in Debug/Profile  (not monolithic builds) that why they used version from CrySystem instead of Launcher.
LmbrWaf patched Launcher, so we could use same code path for monolithic and not build on windows


ranilianaws:
Hello @Ximura,
We are currently reviewing the submission and we will get back to you as soon as possible.

ranilianaws:
Hello @Ximura,

We have accepted the code submission and will include it in a future release of Lumberyard. 

Thank you again for contributing to Lumberyard!

PR #29, opened 2017-11-28
set version in session search

Ximura:
In case server and client has not default value in gm_version cvar (not 1), client couldn't find and connect to server

Client always search for server with version **1**, even when his gm_version is different

I suggest use VF_CONST_CVAR for **gm_version** to prevent client from changing it (ex. by adding it to config)


dev/Gems/Multiplayer/Code/Source/MultiplayerCVars.cpp
```diff
@@ -588,7 +588,7 @@ namespace Multiplayer
         carrierDesc.m_enableDisconnectDetection = !!gEnv->pConsole->GetCVar("gm_disconnectDetection")->GetIVal();
         carrierDesc.m_connectionTimeoutMS = 10000;
         carrierDesc.m_threadUpdateTimeMS = 30;
-        carrierDesc.m_version = gEnv->pConsole->GetCVar("gm_version")->GetIVal();
```
Ximura: we already set version in Multiplayer::Utils::InitCarrierDesc

AMZN-puvvadar:
Hi @Ximura, 

Would you be able to resubmit your diff? The current diff appears to have pulled in a multitude of irrelevant files (css changes etc.) that I imagine were not intended to be part of the change. In its current state the diff cannot be pulled in for review. Thanks!

Ximura:
Hi @AMZN-puvvadar 
You can check diff here
https://github.com/fraglab/Lumberyard/commit/b40a9d4c0b74598a2be806148d999d44121cf248

AMZN-puvvadar:
Hi @Ximura,

I'm pleased to report that we've accepted this changelist. Thanks for contributing to Lumberyard!

PR #31, opened 2017-11-29
Fixed specular map chromaticity

alex-leleka:
Fixed specular map chromaticity of Illum POM materials. mdSelfShadowStrength never used in Pixel shader.
@Amazon please consider adding shader files to github.


alex-leleka:
SelfShadowStrength is actually used

PR #33, opened 2017-12-24
Update values when changed for mesh asset and material override

nocturnals:
Issue:
When you change "Mesh asset" or "Material override" in the mesh component in the editor the strings don't update and shows the previous value.

Reproduce: 
Select an object (mesh)
Change material override. 
Change it again. 
-> string doesn't update (shows 'old' value in the editor)

(Yeah, I know... small bug :-)


ranilianaws:
Hello @nocturnals,

This issue has already been addressed internally and will be included in a future version. We would like to thank you for your contribution!

Thanks!

PR #35, opened 2018-01-01
Added vertex scope to alembic importer (blender export)

nocturnals:
Background:
I was trying out the geometry cache component and I found out that it wouldn't process an animation I made in blender that I tried to export as an alembic file.

Issue:
The alembic part of the resource compiler only supported face-varying indexing while blender can export it as vertex indexing (Alembic::AbcGeom::kVertexScope vs Alembic::AbcGeom::kFacevaryingScope)


ranilianaws:
Hello @nocturnals,

Did you still want to get this code evaluated or would you rather we create a feature request instead? 
Thank you!

nocturnals:
I've opened a new PR. I think it's stable now. Until a few days ago I was unfamiliar with alembic and thought there would be a lot more issues/bugs with the code.
So, yes, please evaluate the code and see if there are other issues/bugs I missed.

PR #36, opened 2018-01-04
Alembic vertex geoscope

nocturnals:
second attempt: It seems stable now. I've been trying out several blend file I downloaded and it all seems to import correctly. If you could test max & maya imports on your side to make sure they still work. Should've left the original pr open but thought there would be more issues.

If you come across a bug, post the .abc file somewhere and I'll debug it. I'd rather have you expend your resources to improve other parts of the engine.

(original pr)
Background:
I was trying out the geometry cache component and I found out that it wouldn't process an animation I made in blender that I tried to export as an alembic file.

Issue:
The alembic part of the resource compiler only supported face-varying indexing while blender can export it as vertex indexing (Alembic::AbcGeom::kVertexScope vs Alembic::AbcGeom::kFacevaryingScope)


ranilianaws:
Hello @nocturnals,

Thank you for your contribution. We have reviewed the code and will include it in a future release.

PR #39, opened 2018-01-16
Test mode for localizaiton (sys_localization_test)

ximura:
Add sys_localization_test command for localization testing

sys_localization_test - provides mechanism to render localization token instead of text. It provides mechanism to check that all text are localizable


gFunc:
Thanks so much for the new localization command.  I'll be looking into pulling this and will keep you posted on any updates :)

gFunc:
Hi, @ximura!
Sorry for the delay!  Your pull request will be included in the next version of the Lumberyard engine. Thanks for contributing to Lumberyard!

PR #41, opened 2018-01-17
parse argc/argv

ximura:
Currently argc and argv in AzFramework::Application  is always initialized with default values.

pass argc/argv from launcher (main) to Application, so that AzFramework::CommandLine could parse correct values.

### DIFF
--- a/dev/Code/Framework/AzFramework/AzFramework/Application/Application.cpp
+++ b/dev/Code/Framework/AzFramework/AzFramework/Application/Application.cpp
@@ -145,11 +145,6 @@ namespace AzFramework
         }
     }
     
-    Application::Application()
-    : Application(nullptr, nullptr)
-    {
-    }
-    
     Application::Application(int* argc, char*** argv)
         : m_appRootInitialized(false)
     {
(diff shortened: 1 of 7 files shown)


dev/Code/Framework/AzFramework/AzFramework/Application/Application.cpp
```diff
@@ -317,11 +311,6 @@ namespace AzFramework
         {
             m_commandLine.Parse(*m_argC, *m_argV);
         }
-        else
-        {
-            m_commandLine.Parse();
```
Ximura: in current implementation m_argC and m_argV is always initialized (line 162 - line 172), so this code create a false assumption.

ranilianaws: Not sure I really follow how the arguments are always initialized? I can see the changes made to the Launchers might be useful but I don't see the need for the Application changes. Do you agree?

ranilianaws:
Hello @ximura,
Unfortunately, we can not accept the code as is due to issues with using __argc and __argv when it comes to other platforms and launchers. Thank you for contribution.

PR #42, opened 2018-01-17
Fixed mesh/brush physics proxy conversion

SSPkrolik:
This worked fine in 1.11, and was broken as of 1.12.
Simply returned code from previous version.

### DIFF
--- a/dev/Code/Sandbox/Editor/LegacyEntityConversion/LegacyEntityConversion.cpp
+++ b/dev/Code/Sandbox/Editor/LegacyEntityConversion/LegacyEntityConversion.cpp
@@ -154,6 +154,13 @@ namespace LegacyConversionInternal
             meshComponentId
         };
 
+        AZ::Uuid editorStaticPhysicsComponentId = "{C8D8C366-F7B7-42F6-8B86-E58FFF4AF984}";
+        if (physics)
+        {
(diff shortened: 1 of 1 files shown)


AMZN-puvvadar:
Hi @SSPkrolik ,

Thanks for the submission! We've begun evaluating it and will update with any responses/status changes. Thanks again!

AMZN-puvvadar:
Hi @SSPkrolik ,

Can you provide repro steps/a use case for the issue this is resolving? CryPhysics entity conversion is currently not supported so we'd like to better understand the motivation behind the change.

SSPkrolik:
Hi, @AMZN-puvvadar well, actually, it was supported in legacy entity upgrade in 1.11 but is missing in 1.12, so basically when you convert physicalized Brush to Component Enity you get only mesh component without static physics and mesh collider on it. So we had to add it manually which is quite a mess for hundreds of objects on level.

AMZN-puvvadar:
Hi @SSPkrolik ,

Thanks for clarifying! I'll deliver that feedback and will update when I have more information.

AMZN-puvvadar:
Hi @SSPkrolik ,

We understand the meat of the issue and instead of taking this pull request as is, we will aim to implement a more final fix for the specific use case. Thanks for your contribution!

PR #43, opened 2018-01-18
GetChildByName method in TransformComponent

LatzekLat:
GetChildByName method is required to be able to manipulate entities inside slice(which has script canvas) if slice was spawned dynamically, and there is no possibility to link in advance(unlike static entities on level)


ranilianaws:
Hello @LatzekLat,

Thank you for your contribution. We are actively reviewing the code and will get back to you on status updates.

Thanks again!

ranilianaws:
Hello @LatzekLat,

We have reviewed the code and have decided against taking it as we do not encourage changes to the core API that rely on Entity Name uniqueness logic and decided to leave it up to the developer to create helper functions on their own. 

Thank you again for contributing to Lumberyard!

PR #44, opened 2018-01-26
1 12 color picker fix

SSPkrolik:
Currently edititing color in hue boxes (on top of color picker) does not send proper signal to Particle Editor to change color.
This fixes the problem


gFunc:
Nice pull request!  The team is evaluating; I'll keep you updated :)

AMZN-alexpete:
This fix is included in version 1.24, Thanks again!

PR #45, opened 2018-01-30
Created Marshalers gem and localized all net marshaling code here.

tolya7:
Implemented new marshalers based on quantization algorithms to allow to get enough precision with lower bandwidth.
Implemented string marshaler based on string table to send each string ones per channel. Marshal only string id after this.
Added ReadBuffer::ReadRawBits to allow reading N bits and trimming unnecessary bits


AMZN-puvvadar:
Hi @tolya7 ,

Thanks for your submission! We have begun evaluating it and will update here with any questions or status changes.

AMZN-puvvadar:
Hi @tolya7 ,

We're really curious how you are binding to the marshalers from the gem. Our assumption is that you are using an application defined component that binds to the transform through an eBus. We would really like to know more about your binding method and if you have any concerns about performance related to this method.

Thanks!

tolya7:
Hi @AMZN-puvvadar ,

To see marshalers performance execute:
mphost
CmdMarshalersTest

Marshaling using StringTable is expensive on Linux:
average time: 3.5 mcs (0.2 mcs on my PC);
max time: 4000 mcs (100 mcs on my PC).

The same problem with any container marshaling.

Thanks.

PR #46, opened 2018-02-15
Fix shifted shadows if Total illumination is Active and e_GI = 0

alex-leleka:
In editor enabling Total illumination shifts shadows even when GI is disabled.


AMZN-alexpete:
Hi @alex-leleka, thank you for submitting this fix.  We are currently reviewing it.

AMZN-alexpete:
Greetings @alex-leleka!  Thank you again for submitting this fix, it will be included in the next release of Lumberyard.

PR #47, opened 2018-02-15
Fix NANs produced by particles when Facing is set to Free

alex-leleka:
NANs produced particles shader when Facing is set to Free. Setting Aspect to 0 makes basis vector vX == 0.


gFunc:
Nice find and fix @alex-leleka!  I'll put this change up for review, and let you know if it's taken into Lumberyard proper.
Thanks!

alex-leleka:
Hello @gFunc I've created a post on the forum https://gamedev.amazon.com/forums/articles/68210/fix-nans-produced-by-particles-when-facing-is-set.html

PR #49, opened 2018-03-01
NetBindingSliceInstantiationHandler

ximura:
fix crash in NetBindingSliceInstantiationHandler, after disconnect from dedicated server before all assets was instantiated on remotes


AMZN-puvvadar:
Hi @ximura ,

Thanks for the submission, we've begun reviewing this. It seems like a very straightforward change so I hope to have a response on this shortly.

AMZN-puvvadar:
Hi @ximura,

We've accepted this change into Lumberyard, thanks again for the contribution!

PR #50, opened 2018-03-12
Drawing tablet support

nocturnals:
I've added drawing tablet support. At the moment I've only converted two edit tools: TextureTerrainPainter and VegetationTool. Pen pressure only.
I wanted to get some feedback in the mean time so when I upgrade other tools, I can implement them correctly using your suggestions.


ranilianaws:
Hello @nocturnals,

Thank you for the contribution. We will evaluate the code and get back to you as soon as possible.

PR #52, opened 2018-03-14
Fix for Lua netSynced properties.

tkgdhughes:
# Lua NetSynced Properties Fix

### Description
Lua netSynced properties were initialized on first dataset update, that means if a dataset changed value infrequently (i.e. its value was set once on server in OnActivate() to a different value than the one in the properties section in the script), the clients would never see it.

ScriptComponentReplicaChunk::AssignDataSetIfAvailable(const AZStd::string& valueName, ScriptNetBindingTable::NetworkedTableValue& tableValue) is called from the Proxy when the script chunk is assigned to ScriptNetBindingTable, and when a new property is registered.
This will initialize NetworkedTableValue with the current property (if there is one already available).
This ensures that we start using proper, not shimmed, properties as soon as they are available rather than after they have changed at least once.
Without this fix, if a property doesn't change after a client connected to the server, the client will never use the actual property value, and happily use the shimmed property value instead.

### Tested against
LY 1.12 D.R.G. Initiative.

### DIFF
--- a/dev/Code/Framework/AzFramework/AzFramework/Script/ScriptComponent.cpp
+++ b/dev/Code/Framework/AzFramework/AzFramework/Script/ScriptComponent.cpp
@@ -671,6 +671,13 @@ namespace AzFramework
             lua_pop(lua, 1); // remove the OnActivate result
         }
 
+        if (m_netBindingTable)
+        {
+            // Set the flag indicating that the owning entity is active. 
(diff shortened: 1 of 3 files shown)


tkgdhughes:
Thanks, @tkgszaluga, I'm sure that will help the verification process!

tkgdhughes:

PR #54, opened 2018-03-14
9) Fix for allocator double creation

tkgdhughes:
# Don't Double Create Allocators

### Description
Prevent the double creation of memory allocators. 
This was causing an assert in the unit tests which could leave the build running forever.

### DIFF
--- a/dev/Code/Framework/AzCore/AzCore/Memory/Memory.h
+++ b/dev/Code/Framework/AzCore/AzCore/Memory/Memory.h
@@ -381,14 +381,16 @@ namespace AZ
             if (!s_allocator)
             {
                 s_allocator = Environment::CreateVariable<Allocator>(AzTypeInfo<Allocator>::Name());
-                if (s_allocator->IsReady()) // already created in a different module
-                {
-                    return;
-                }
             }
             else
             {
                 AZ_Assert(s_allocator->IsReady(), "Allocator '%s' already created!", AzTypeInfo<Allocator>::Name());
+            }
+
+            // already created, do not create again
+            if (s_allocator->IsReady()) 
+            {
+                return;
             }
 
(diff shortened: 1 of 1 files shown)


ranilianaws:
Hello @tkgdhughes ,

Thank you for the contribution! We are evaluating the code and will get back to you as soon as possible!

tkgdhughes:

ranilianaws:
Hello @tkgdhughes ,

We reviewed the code and decided against taking it as it breaks the expected use case of the system. We are curious however about the particular use case and how the system is being used. Could you share more information regarding the unit tests and how they are being executed?

tkgdhughes:
Hi @ranilianaws,

Thanks for the feedback. I am curious to know why this breaks the expected use case? 

I may be misinterpreting the code but it appears that if **s_allocator** is not null, and **->IsReady()** returns true, the system will assert _("Allocator '%s' already created!")_ but still continue to **->Create(desc)**. The previous behaviour had the function early out when **->IsReady()** returns true when the allocator has already been created in a different module.

I will look into providing more information on the unit tests shortly as I am just heading out of the office now and only have a half day tomorrow.

Have a good day!
Dan

ranilianaws:
Hello @tkgdhughes ,

We believe that you might have discovered a root cause or use case that we might not be aware of. We expect that when the Create() function is called more than once with the same descriptor to fail as it should only be called once. In order to prevent that from happening, you should call isReady on the AllocatorInstance first before calling Create if needed.

Hope that helps.
Have a great night!

tkgdhughes:
I suppose I'm just a little confused since the original behaviour is still present but this is just extra safety to also catch the case where the allocator is not null on entry in non-debug builds.

(I'm back online because my night has ended up doing more programming!)

PR #56, opened 2018-03-15
38) Add "Collapse All" button to the Property Editor UI

tkgdhughes:
# Add "Collapse All" Button To Editor Property UI

### Description
This is actually one of the simplest, yet one of my favorite, additions to the Lumberyard UI; adding a "Collapse All" button to the UI. 

The "Collapse All" button is incredibly useful when working on a component entity with a large number of components on it. For example, our player character has around thirty components on it. This means that to find the component you wish to change, there is a lot of scrolling required and even when you find the correct component there is a lot of visual clutter. If you wish to edit a couple of components at once, being able to collapse all but the ones you are interested in reduces the amount of excess noise.

Admittedly this modification is a bit subjective but I can confirm that our lead designer approves of this feature and uses it over the search bar!

Hope you like it.

### Check it out in action
expandallgif

### DIFF
--- a/dev/Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/EntityPropertyEditor.cpp
+++ b/dev/Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/EntityPropertyEditor.cpp
@@ -1100,12 +1100,9 @@ namespace AzToolsFramework
     {
         if (m_gui && m_gui->m_collapseComponentsButton)
         {
-            static const char* collapseAllText = "Collapse All";
-            static const char* expandAllText = "Expand All";
-
             m_componentsExpanded = !m_componentsExpanded;
 
-            m_gui->m_collapseComponentsButton->setText(m_componentsExpanded ? collapseAllText : expandAllText);
+            m_gui->m_collapseComponentsButton->setText(m_componentsExpanded ? tr("Collapse All") : tr("Expand All"));
             for (auto componentEditor : m_componentEditors)
             {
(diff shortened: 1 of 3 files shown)


nocturnals:
I like the addition!

I'm new to QT so what I'm about to say could be completely wrong (and in any case need to check my own commits because I forgot about translations)

Translations:
In EntityPropertyEditor.cpp line 1103:
```
           static const char* collapseAllText = "Collapse All";
           static const char* expandAllText = "Expand All";
....
            m_gui->m_collapseComponentsButton->setText(m_componentsExpanded ? collapseAllText : expandAllText);        
```
Would a setText(...) -> setText(tr(...)) solve it?

tkgdhughes:
Hey @nocturnals,

Thanks for the feedback, translations had actually slipped my mind too! Doing a quick Google suggests that tr(...) would work. It also looks like that LY codebase uses it extensively too. I'll drum up that change shortly. 

Thanks!

tkgdhughes:

ranilianaws:
Hello @tkgdhughes ,

Thank you for the contribution. We will evaluate the code and get back to you.

ranilianaws:
Hello @tkgdhughes ,

We made slight modifications to the code and removed the button but left the context menu item as it is more consistent with our UI/UX standards. In the future, we will be making more UI/UX changes to the Entity Inspector and will revisit how we can collapse/expand components again. 

Thank you for the contribution!

tkgdhughes:
Good stuff! Interestingly I had actually started with it on a context menu item but then changed it. I should have stuck with my initial change!
Thanks

PR #58, opened 2018-03-15
2) Fix for AssetProcessorManager thread-safety

tkgdhughes:
# AssetProcessorManager::IsIdle is not thread-safe

### Description
Modified AssetProcessorManager::IsIdle to make it (more) thread-safe and mitigate against un-defined behavior.

IsIdle() is normally called from the AssetProcessorManager's worker thread only, however, in batch mode, it is also called from the main thread, making m_filesToExamine.isEmpty(), m_activeFiles.isEmpty(), etc. not safe.
The solution presented is not pretty, but it avoids locking when IsIdle() is called from the worker, which occurs pretty much constantly, whereas IsIdle() on the main thread is called infrequently, only in batch mode, when OnBecameIdle signal is emitted.

### Tested against
LY 1.12 StarterGame,
LY 1.12 D.R.G. Initiative.

### DIFF
--- a/dev/Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
+++ b/dev/Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
@@ -2420,13 +2420,39 @@ namespace AssetProcessor
 
     bool AssetProcessorManager::IsIdle()
     {
-        if ((!m_queuedExamination) && (m_filesToExamine.isEmpty()) && (m_activeFiles.isEmpty()) && 
-            !m_processedQueued && m_assetProcessedList.empty() && m_jobsToProcessLater.empty())
+        // 	Modified this function to make it (more) thread-safe. 
+		//	IsIdle() is normally called from the AssetProcessorManager's worker thread only.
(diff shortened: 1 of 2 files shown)


ranilianaws:
Hello @tkgdhughes ,

Thank you for your contribution! We will review the code and get back to you as soon as possible.

tkgdhughes:

ranilianaws:
Hello @tkgdhughes ,

I don't think I see where OnBecameIdle comes from. Which file does this refer to?

Also can you provide a test case we can reproduce this under? We do not expect isIdle to be called from another thread.

Thank you!

tkgdhughes:
@ranilianaws, my apologies, it should be Q_EMIT BecameIdle();, not OnBecameIdle, from rccontroller.cpp

ranilianaws:
Hello @tkgdhughes ,

Thank you for the contribution. We have fixed this issue taking a different approach and as such we will not be taking this code.

tkgdhughes:
Hey @ranilianaws, what approach did you guys take for this issue?

ranilianaws:
Hey @tkgdhughes ,

We made some major changes to how the AP works. It is now event-based using QT slots and will no longer require thread locking as the events themselves are atomic.

tkgdhughes:
Thanks for the update!

PR #65, opened 2018-03-20
Fix For Lua Net Bindings Crash

tkgdhughes:
### Description
Fix for Lua net bindings trying to delete their datasets multiple times, which leads to a crash - datasets bound to a network table value should transfer the script property ownership to the reserver instead of deleting it. If there is no network table value bound to the dataset delete it as before.


tkgdhughes:

AMZN-alexpete:
Hi @tkgdhughes!  Thanks for submitting this fix, we're reviewing the change now.

AMZN-alexpete:
Hi @tkgdhughes can you provide repro steps for this issue?

PR #67, opened 2018-03-20
4) Add Dynamic Slice Network Cloning

tkgdhughes:
### Description
Added cloning dynamic slices support to game entity context bus.

### Usage

```c++
        AZ::EntityId ExampleInventoryComponent::CloneItem(AZ::EntityId itemId, AZ::EntityId parentId)
        {
        // 	If it is an item from a slice, clone the entire slice. 
	//	That way the ancestry info will be preserved.
            if (!parentId.IsValid())
            {
                AZ::SliceComponent::SliceInstanceAddress sliceAddr;
                EBUS_EVENT_ID_RESULT(sliceAddr, itemId, AzFramework::EntityIdContextQueryBus, GetOwningSlice);
                if (sliceAddr.first && sliceAddr.second)
                {
                    // This is a slice entity. Clone the entire slice
                    AZ::SliceComponent::EntityIdToEntityIdMap entityIdMap;
                    AZ::SliceComponent::SliceInstanceAddress clonedSliceAddr;
                    EBUS_EVENT_RESULT(clonedSliceAddr, AzFramework::GameEntityContextRequestBus, CloneDynamicSliceByEntity, itemId, entityIdMap);
                    if (clonedSliceAddr.first && clonedSliceAddr.second)
                    {
                        auto clonedItemId = entityIdMap[itemId];
                        return clonedItemId;
                    }
                    else
                    {
                        // We were unable to clone the slice. In theory we could just fall back to cloning entities one by one,
                        // but that would hide the bug that caused the cloning operation to fail.
                        AZ_Error("ExampleInventoryComponent", false, "Failed to clone dynamic slice");
                        return AZ::EntityId();
                    }
                }
                // else - Not a slice entity. Fall back to cloning entities one by one
            }

            // Clone entities one by one recursively.
            // ...

            return clonedItemId;
        }
```

### DIFF
--- a/dev/Code/Framework/AzFramework/AzFramework/Entity/GameEntityContextBus.h
+++ b/dev/Code/Framework/AzFramework/AzFramework/Entity/GameEntityContextBus.h
@@ -123,6 +123,14 @@ namespace AzFramework
         virtual bool DestroyDynamicSliceByEntity(const AZ::EntityId& /*id*/) = 0;
         
         /**
+        * Clone an entire dynamic slice instance given the ID of any entity within the slice.
+        * @param entityId The ID of the entity whose dynamic slice instance you want to clone.
+        * @param entityIdMap A reference to a map that will be filled with source->copy entity ID mappings.
+        * @return Cloned slice address if the dynamic slice instance was successfully cloned, null address otherwise.
+        */
+        virtual AZ::SliceComponent::SliceInstanceAddress CloneDynamicSliceByEntity(const AZ::EntityId& /*entityId*/, AZ::SliceComponent::EntityIdToEntityIdMap& /*entityIdMap*/) { return AZ::SliceComponent::SliceInstanceAddress(); }
+
+        /**
          * Instantiates a dynamic slice asynchronously.
          * @param sliceAsset A reference to the slice asset data.
          * @param worldTransform A reference to the world transform to apply to the slice. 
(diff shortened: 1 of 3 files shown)


AMZN-alexpete:
Hi @tkgdhughes, thanks for submitting this feature.  Question, why does the title say network cloning?  I didn't see something specific to networking in here - it seems fairly generic.

tkgdhughes:
Good morning @AMZN-alexpete,

You are correct, it is fairly generic. The title is left over from the initial problem which we were addressing. We were having issues with custom metadata for inventory items not being found for network clients. By implementing CloneDynamicSliceByEntity we were able to fix this issue. I'll remove "Network" from the title!

Thanks,
Dan

AMZN-alexpete:
Hi @tkgdhughes can you provide more details about the specific use cases for this please?

It looks like this function creates a copy of entities already in the world. Even though they came from a slice, this preserves the local modifications of the live entities.  Does it matter that those entities came from a slice? Would it be more useful to just provide a function that can clone any entities, whether or not they came from a slice? Something like:

// Returns ID of cloned entity
AZ::EntityId CloneEntityAndDescendants(AZ::EntityId entity);

There is also an existing method SliceComponent::SliceReference::CloneInstance() that clones slice entities, that could be used but it's not clear what problem is solved by maintaining a connection to the original slice at runtime.

Thanks and sorry for the delay!

PR #68, opened 2018-03-21
32) Fix D3D renderer crash

tkgdhughes:
### Description 

Safety check against DBZ in CD3D9Renderer::CheckSSAAChange where width/height came through as zero, potentially due to cvar problems which we didn't get to the bottom of but could at least protect against.


AMZN-alexpete:
Hi @tkgdhughes, thank you for submitting this fix, it will be including in an upcoming version of Lumberyard.

PR #69, opened 2018-03-21
NetBindingSliceInstantiationHandler

ximura:
fix crash in NetBindingSliceInstantiationHandler after disconnect from dedicated server before all assets was instantiated on remotes
in case asset was already added to TickBus in EntityContext::OnAssetReady


ranilianaws:
Hello @ximura,

Thank you for the contribution! We will review the code and get back to you soon!

ranilianaws:
Hello @ximura,

Can you provide more context as to how to reproduce the mentioned crash?

ranilianaws:
I see, thank you for the additional context @ximura. Upon review we discovered that we had an alternative fix for this in a future release. The fix is to capture the variable passed into the Lambda function as such
[bindingQueue=m_bindingQueue] instead of [this].

PR #70, opened 2018-03-21
11) Fix editor silent crash

tkgdhughes:
### Description
Fix for the editor silently crashing when the bin folder is named bin64vc140 instead of Bin64vc140 (upper/lower case 'B').

Replaced regex-replace string modification with a simpler one (endsWith() + chop()).

There were two issues with the regex:
1) It is case-sensitive and failed when the executable path was using lower case (i.e. app path used bin32 instead on Bin32)
2) The regex would replace all occurrences of Bin32/Bin64 instead of only the one at the end of the path.

### Note 

On testing it looks like the silent crash initially discovered appears to have been fixed, however, when single-stepping ::InitDirectory and ::SetMasterCDFolder it is trivial to see that the regex code is flawed as described above and should be replaced with the simpler and more correct version.

### DIFF
--- a/dev/Code/Sandbox/Editor/CryEdit.cpp
+++ b/dev/Code/Sandbox/Editor/CryEdit.cpp
@@ -2016,11 +2016,24 @@ void CCryEditApp::InitDirectory()
     QString szExeFileName = qApp->applicationDirPath();
 
     // Remove Bin32/Bin64 folder/
-    szExeFileName.remove(QRegularExpression(R"((\\|/)Bin32.*)"));
-
-    const QString bin = QRegularExpression::escape(BINFOLDER_NAME);
-    const QString reg = QStringLiteral("(\\\\|/)%1.*").arg(bin);
-    szExeFileName.remove(QRegularExpression(reg));
+    // Replaced regex-replace string modification with a simpler one (endsWith() + chop()).
+    // There were two issues with the regex:
+    //  1) It was case-sensitive, and failed when executable path was using lower case(i.e. app path used bin32 instead on Bin32)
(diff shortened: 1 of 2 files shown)


ranilianaws:
Hey @tkgdhughes ,

ranilianaws:
Hello @tkgdhughes ,

Upon evaluation of the code we were wondering if the following would work instead:


```cpp
const QString bin = QRegularExpression::escape(BINFOLDER_NAME);
const QString reg = QStringLiteral("(\\\\|/)%1(?!.*%1).*$").arg(bin);
szExeFileName.remove(QRegularExpression(reg, QRegularExpression::CaseInsensitiveOption));
```

That provides case sensitive and also only removes the last instance in the string instead of all of them (because of the ?! negative group and the $ indicating the end of line). Also, we don't need to worry about Bin32 anymore, in the Editor at least.

tkgdhughes:
Hey @ranilianaws,

Good regex skills! I would still argue that the endsWith() + chop() approach is easier to understand and easier to support if any changes are needed in the future. I know regex, but I still need to think hard to understand that expression! If someone without a good versing in regex needs to debug this they will need to do some googling whereas a simple chop is easy to understand.

Thoughts?
Dan

ranilianaws:
Hey @tkgdhughes ,

Agreed the provided code is easier to read however endsWith might not work across multiple platforms where the path would be Bin64/something/something. We will adjust the regex and add comments to help clarify what it's attempting to do. Thank you again for the submission. :)

tkgdhughes:
No worries, thanks for the feedback!

PR #71, opened 2018-03-22
12) Fix for crashes and hangs during level loading

tkgdhughes:
### Description

We were getting a lot of crashes and hangs while loading into our main level. Our main level was massive and took a good while to load everything in. One of the things we made extensive use of in the level was rendering canvases to textures. We narrowed down our crashes and hangs on loading to the fact that ClearDepthStencilView was being called from the main thread by UiCanvasComponent::CreateRenderTarget. This was corrupting the D3D context, resulting in missing fences (hang) or GPU corruption (crashes).

The fix for this is to not issue a ClearDepthStencilView from FX_CreateDepthSurface; there should be no reason that a depth target needs to be cleared upon creation, as it should be cleared or filled every frame.


AMZN-alexpete:
Hi @tkgdhughes ,  we have integrated this fix, and modified the code at a higher level to prevent FX_CreateDepthSurface from being used on the main thread.  This change will be included in an upcoming version of Lumberyard.  Cheers!

PR #72, opened 2018-03-22
17) Fix for ambiguous entity id compile issue

tkgdhughes:
### Description

Simple one here; fix for Linux uber build where a "using namespace AZ" bleeds out to other CPPs and causes an ambiguous EntityId issue (As compiler cannot choose between the old cry EntityId or the newer AZ:EntityId).


AMZN-alexpete:
Thanks for this contribution @tkgdhughes.  We will be including this fix, with some minor modifications, in an upcoming version of Lumberyard

PR #74, opened 2018-03-22
24) Fix for Lua enums

tkgdhughes:
### Description
Fix the string combo box handler to handle enumValue attributes. These use pair<string,string> in Lua.

### Usage
Put the following code into a Lua script and test before and after this change. Lua enum properties should now work correctly.

```c++	
local EnumTest = 
{
	Properties = 
	{
		EnumType = 		
		{
    		default = "DefaultType",
    		ui = "ComboBox",
			enumValues =	
			{
				{ "1", "DefaultType" },
				{ "2", "OtherType" },
			}
		},
	},
}

function EnumTest:OnActivate()
	Debug.Log(self.Properties.EnumType)
end

return EnumTest
```

### DIFF
--- a/dev/Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyStringComboBoxCtrl.cpp
+++ b/dev/Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyStringComboBoxCtrl.cpp
@@ -86,6 +86,21 @@ namespace AzToolsFramework
         m_pComboBox->blockSignals(false);
     }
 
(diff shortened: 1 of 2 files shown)


ranilianaws:
Hey @tkgdhughes ,

Thank you for the contribution! We will review the code and get back to you soon!

ranilianaws:
Hello @tkgdhughes ,

Unfortunately, we will not be taking this pull request as we have discovered during the review that there is already a system coming online soon for a Generic ComboBoxCtrl that allows for a combo box to be constructed with a mapping of any C++ type to an AZStd::string instead.

Thank you again for the contribution.

PR #76, opened 2018-03-22
33) Fix for render thread stall

tkgdhughes:
### Description
Fix for render thread stall caused by the use of an inefficient condition variable being used for lock-stepping. The render thread was often not waking for several milliseconds after the main thread signalled it, causing a CPU bottleneck. Fixed by changing it to use an Event.

Upped the main thread and render thread priorities to ABOVE_NORMAL. The reasoning is that these threads drive the job threads, so we don't want them delaying if they have something to do. By putting them at the same priority, Windows should favour putting them each on their own cores, with the job threads moving on to the spare cores.

### DIFF
--- a/dev/Code/CryEngine/RenderDll/Common/RenderThread.cpp
+++ b/dev/Code/CryEngine/RenderDll/Common/RenderThread.cpp
@@ -62,6 +62,7 @@ void CRenderThread::Run()
     CNameTableR::m_nRenderThread = renderThreadId;
     gEnv->pCryPak->SetRenderThreadId(renderThreadId);
     //SetThreadAffinityMask(GetCurrentThread(), 2);
+    SetThreadPriority(GetCurrentThread(), THREAD_PRIORITY_ABOVE_NORMAL);
     m_started.Set();
     gRenDev->m_pRT->Process();
     gEnv->pSystem->GetIThreadTaskManager()->MarkThisThreadForDebugging(RENDER_THREAD_NAME, false);
@@ -120,6 +121,7 @@ SRenderThread::SRenderThread()
     m_bBeginFrameCalled = false;
     m_bQuitLoading = false;
 #if defined(USE_HANDLE_FOR_FINAL_FLUSH_SYNC)
+    m_FlushCondition = CreateEvent(NULL, FALSE, FALSE, "FlushCondition");
     m_FlushFinishedCondition = CreateEvent(NULL, FALSE, FALSE, "FlushFinishedCondition");
 #endif
     Init(2);
@@ -181,6 +183,7 @@ SRenderThread::~SRenderThread()
     QuitRenderLoadingThread();
     QuitRenderThread();
 #if defined(USE_HANDLE_FOR_FINAL_FLUSH_SYNC)
+    CloseHandle(m_FlushCondition);
     CloseHandle(m_FlushFinishedCondition);
 #endif
 }
@@ -3355,7 +3358,7 @@ void SRenderThread::FlushAndWait()
 // Flush current frame without waiting (should be called from main thread)
 void SRenderThread::SyncMainWithRender()
 {
-    AZ_PROFILE_FUNCTION_STALL(AZ::Debug::ProfileCategory::Renderer);
+    AZ_PROFILE_FUNCTION(AZ::Debug::ProfileCategory::Renderer);
     FUNCTION_PROFILER_LEGACYONLY(GetISystem(), PROFILE_RENDERER);
 
 #if defined(USE_HANDLE_FOR_FINAL_FLUSH_SYNC)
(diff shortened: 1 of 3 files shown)


dev/Code/CryEngine/RenderDll/Common/RenderThread.h
```diff
@@ -234,10 +234,11 @@ struct SRenderThread
 #ifdef USE_LOCKS_FOR_FLUSH_SYNC
     int m_nFlush;
     CryMutex                            m_LockFlushNotify;
-    CryConditionVariable    m_FlushCondition;
 #if defined(USE_HANDLE_FOR_FINAL_FLUSH_SYNC)
+    HANDLE m_FlushCondition;
     HANDLE m_FlushFinishedCondition;
 #else
+    CryConditionVariable    m_FlushCondition;
```
tkgdhughes: We were seeing cases where the render thread was not waking immediately after the main thread signalled it. Previously this was using a heavy-weight condition variable, which was overkill and causing the inefficiency. Changed it to use a lightweight Event instead.

dev/Code/CryEngine/RenderDll/Common/RenderThread.cpp
```diff
@@ -3355,7 +3358,7 @@ void SRenderThread::FlushAndWait()
 // Flush current frame without waiting (should be called from main thread)
 void SRenderThread::SyncMainWithRender()
 {
-    AZ_PROFILE_FUNCTION_STALL(AZ::Debug::ProfileCategory::Renderer);
+    AZ_PROFILE_FUNCTION(AZ::Debug::ProfileCategory::Renderer);
```
tkgdhughes: This is not a stall function. It handles the buffer swaps and render thread submission.

dev/Code/CryEngine/RenderDll/Common/RenderThread.cpp
```diff
@@ -3463,6 +3466,8 @@ bool SRenderThread::IsFailed()
 bool CRenderer::FlushRTCommands(bool bWait, bool bImmediatelly, bool bForce)
 {
     AZ_TRACE_METHOD();
+    AZ_PROFILE_FUNCTION_STALL(AZ::Debug::ProfileCategory::Renderer);
```
tkgdhughes: Added a profile marker so the main thread stalling on the render thread is visible in the profiler.

tkggwatson:
Just wanted to add to Dan's description here - this change made a difference of about 20fps in some areas of our map.

ranilianaws:
Hey @tkgdhughes & @tkggwatson ,

Thank you for the submission. We will review the code and get back to you soon.

ranilianaws:
Hello @tkgdhughes & @tkggwatson ,

The threading performance issue definitely exists. We were wondering if you have tried to use a spin_mutex or shared_mutex instead of changing threading priority (as that could lead to unforeseen circumstances).

tkggwatson:
Hi @ranilianaws,

A spin_mutex would busy-lock the thread, which would be undesirable as it would burn unnecessary cpu cycles that could otherwise be being used by driver and os threads, particular if the render thread completes much quicker than the update thread. Therefore I'd advise against spin_mutex in this case.

I'm not entirely sure what you are proposing with a shared_mutex, as a mutex isn't a signal? I guess you could make it so that the update thread holds the mutex locked during the update, and the render thread attempts to lock it during its wait, but that would just be an obfuscated version of what we have done here I think, so I'm not clear on the benefit.

We chose the Event approach because it's a kernel-level primitive with immediate context-switch by the kernel from within the SetEvent function. It is extremely lightweight in terms of resource usage and create/destroy performance. 

The original condition variable approach is not kernel-level supported, so puts the waiting thread at the mercy of the OS to wake up after the condition is met on the signalling thread. At best this is on the next scheduler quantum. At worst it is many milliseconds later. Hence this is not a good pattern on Windows or XBoxOne for synchronising the main threads governing the application's frame rate.

The addition of raising the thread priority of the critical-path threads is something we've done in most games we've worked on that make heavy use of worker threads. It ensures that if the governing threads have something to do, the kernel will reschedule any other normal-priority threads on to other cores, which helps maintain a consistent frame rate and ensures that workers do not unnecessarily delay the governing threads when they are boosted by the kernel (such as just after their semaphores are raised). Note this is just a thread priority, not a process priority, so the OS is still free to balance other processes - it just tells the schedule what to do when it has multiple choices to wake this processes threads.

Lastly I would note that the sync primitive at the other end of the renderer, m_FlushFinishedCondition, had already previously been converted to an event either by AZ or prior to you adopting the code, so this new change also makes the code more consistent.

From these changes combined, we saw a huge performance increase in several areas of our maps.

Hope that helps validate our thinking!
Graham

nocturnals:
Hi @tkggwatson 

I've noticed a spinlock at dev\Code\CryEngine\CryCommon\IJobManager.h in JobManager::SJobSyncVariable::Wait, SetStopped, SetRunning. What is does are atomic ops and loops them until it unlocks.

I'm mentioning this because I think you used kernel level diag tools but since this is a spinlock, it might not have shown up. If there's no contention, it might actually be faster than doing a kernel transition for thread sync.

tkggwatson:
Hi @nocturnals, 

A spinlock would not be suitable here because the purpose of this piece of code is to lock step the render thread with the main thread, not to protect a section of code. Therefore a primitive that acts as a signal is required. The two threads being lock stepped will compete at different times, and it is undesirable to burn cycles waiting for the other thread to finish.

ranilianaws:
Hello @tkgdhughes ,

We will not take this pull request as the team will be investigating other alternatives to solve this issue. 
Thanks again!

PR #78, opened 2018-03-23
20) Add camera rig update order override

tkgdhughes:
## Description 

By default, TickBus uses TICK_DEFAULT order. Pretty much everything in Lumberyard uses this default order.

If the camera rig listens for events or queries properties of other entities, then it may be ticked:
1) Before all dependent entities, and so it will be behind 1 frame
2) After all dependent entities (usually the desired behaviour)
3) Somewhere in the middle - some dependencies still hold values from the previous frame, others have already been updated.

I said (2) is only usually the desired behaviour, because there are situations where it gets even more complicated, like when the rig depends on entities that use physics or animation. Physics and animation are updated on worker threads, meaning that any update requested in this frame will be applied in the next frame.

For example, if the rig is following a character that uses physics, the character's position/orientation change calculated from this frame's input will be applied in the next frame, but the look direction (calculated from mouse input) is ready in the current frame; if the rig uses it for orientation updates, it will be applied 1 frame too early.

By adding the ability to more tightly control this tick order it is possible to mitigate some of these effects where best suited for the particular game. This was essential for smoothing out aiming and control issues in DRG.

### DIFF
--- a/dev/Gems/CameraFramework/Code/Source/CameraRigComponent.cpp
+++ b/dev/Gems/CameraFramework/Code/Source/CameraRigComponent.cpp
@@ -61,6 +61,8 @@ namespace Camera
 
         m_initialTransform = AZ::Transform::CreateIdentity();
         EBUS_EVENT_ID_RESULT(m_initialTransform, GetEntityId(), AZ::TransformBus, GetWorldTM);
+
+        AZ::TickBus::Handler::m_tickOrder = AZ::TICK_DEFAULT + m_tickOrderOffset;
         AZ::TickBus::Handler::BusConnect();
     }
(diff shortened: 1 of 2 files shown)


AMZN-alexpete:
Hey @tkgdhughes,

swebrett:
Hey, @tkgdhughes .

Thank you for your contribution, but we've decided not to accept it for inclusion in Lumberyard.  TickBus and its order have some issues, like you've illustrated above, and we'd like to take the time to come up with a more general solution for components that may want to update values at different times during a frame, such as both before and after an animation/physics step has completed.  You've given us a lot to think about!

PR #80, opened 2018-03-23
27) Fix for a crash in RenderMesh

tkgdhughes:
### Description

Speculative fix  for a crash in RenderMesh::CreateUVCache() caused by multiple threads resizing a std::vector (speculative as I wasn't able to reproduce the crash locally, so I can't verify if it's fixed. I was, however, able to prove that CreateUVCache() is called from multiple threads, and m_UVCache vector is not protected by any synchronisation primitives, so that's a very likely cause of the crash).

### DIFF
--- a/dev/Code/CryEngine/RenderDll/Common/RenderMesh.cpp
+++ b/dev/Code/CryEngine/RenderDll/Common/RenderMesh.cpp
@@ -367,6 +367,7 @@ int CRenderMesh::Release()
 }
 
 CRenderMesh::CRenderMesh()
+    : m_UVCacheSize(0)
 {
 #if defined(USE_VBIB_PUSH_DOWN)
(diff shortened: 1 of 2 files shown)


AMZN-alexpete:
Hi @tkgdhughes, thank you for submitting this fix.  When investigating this code it looks like the only place that calls CreateUVCache is from CRenderMesh::GetUVPtr which does a SREC_AUTO_LOCK(m_sResLock) in the same block so I'm not sure I understand how adding an additional SREC_AUTO_LOCK inside CreateUVCache will help.

AMZN-alexpete:
Hey @tkgdhughes,

We were not able to reproduce this issue and believe this issue has already been fixed in a more recent version of Lumberyard by removing other locations that were calling into CreateUVCache without acquiring a resource lock.

PR #83, opened 2018-03-26
28) Fix transform spawning inaccuracy

tkgdhughes:
### Description

Fix for entities being spawned at an inaccurate position. 
The inaccuracy was caused by an unnecessary world->local->world transform round trip.

Test by using the SpawnerComponent to spawn an entity at a known transform and log out its transform from OnActivate. There will be a small but significant difference before and after the fix. This was particularly obvious for us when spawning bullet hit decals. The character would aim at a particular spot, raycast to a known position, then the decal would spawn off centre.


ranilianaws:
Hey @tkgdhughes ,

Thank you for the contribution! We will evaluate the code and get back to you soon.

ranilianaws:
Hey @tkgdhughes ,

Thanks for the contribution! This code will make it into a future release of Lumberyard.

PR #84, opened 2018-03-26
29) Fix renderer crash

tkgdhughes:
### Description

Fix for a very elusive renderer crash:
- Add return value safety checks so we don't call D3D functions with a null pointer (should stop the crash). 
- Add protection for MapConstantBuffer returning nullptr.


AMZN-alexpete:
Hi Dan!  If you have any additional information about the circumstances that led to this crash that'd be great to have because we think it fixes the crash but there may still be a root cause that isn't addressed.  In any case, thank you for this fix, it will be included in an upcoming version of Lumberyard!

tkgdhughes:
Thanks @AMZN-alexpete, I'll have a chat with the guys and see what everyone thinks.

PR #85, opened 2018-03-26
24) Fix for final Boot->FE->Exit memory leak.

tkgdhughes:
### Description

Made CryNameR use a static allocated nametable instead of allocating one on first use. This seems to be okay, despite the previous comment saying not to do it this way. I suspect that CryNameR is not used as much these days, as the only breakpoints to hit it seemed to be in shaders.

### DIFF
--- a/dev/Code/CryEngine/RenderDll/Common/CryNameR.h
+++ b/dev/Code/CryEngine/RenderDll/Common/CryNameR.h
@@ -220,17 +220,8 @@ private:
 
     ENGINE_API static CNameTableR* GetNameTable()
     {
-        static CNameTableR* ms_table;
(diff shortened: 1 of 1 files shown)


ranilianaws:
Hello @tkgdhughes ,

Thank you for the submission. We will evaluate the code and get back to you soon!

tkgdhughes:
Not sure to be honest! The simple answer is yes, an alternative could be provided but would need to be worked out. This issue showed up on the DS when memory leak tracking is enabled and the DS is shut down.

ranilianaws:
I will close this pull request as we can't accept the code as it seems too risky to take. Please create another one if needed.

PR #86, opened 2018-03-26
37) Fix for network stall ticker thread disconnection problems

tkgdhughes:
### Description

Disabled Network Stall Ticker Thread, as it is a root cause of a few different crashes. The most recent one seen was caused by running Lua RPC handler from that thread during level loading, causing data corruption.

The Lumberyard team believes it can be safely disabled, as it is a relic that was used by the old CryNetwork system before the GridMate replaced it.
See the related forum ticket: https://gamedev.amazon.com/forums/questions/61765/networkstallticker-thread-unsafe-causing-crashes.html

When the thread is disabled, update GridMate from the main thread during level loading. That ensures that loadLevel request is sent from the server to clients in a timely manner and that the network heartbeat is continuously updated, preventing disconnections during level load.

### DIFF
--- a/dev/Code/CryEngine/CryAction/CryAction.cpp
+++ b/dev/Code/CryEngine/CryAction/CryAction.cpp
@@ -347,6 +347,7 @@ CCryAction::CCryAction()
     , m_pScriptBindDS(0)
     , m_pScriptBindMFX(0)
     , m_pPersistentDebug(0)
+    , m_networkStallTickerReferences(0)
     , m_pMaterialEffectsCVars(0)
     , m_pEnableLoadingScreen(0)
     , m_pShowLanBrowserCVAR(0)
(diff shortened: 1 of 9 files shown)


AMZN-puvvadar:
Hey Dan,

Thanks for the submission! I'll start taking a look at this.

AMZN-puvvadar:
Hey Dan,

One quick note as we're in the review process. The changes made here cause a failure in Unit Tests corrected by updating the signature for SynchronousLoadingTick in ISystemMock.h. I've made this change on our end and wanted to let you know for your own codebase. I'll update again when there's more feedback, cheers!

AMZN-puvvadar:
Hey Dan,

I have a question for you in regards to RenderThread.cpp. In your change you're using your flush callback specifically in the USE_HANDLE_FOR_FINAL_FLUSH_SYNC define check. Is there a reason you don't want that callback to be used in the other logic branches?

tkgdhughes:
Morning @AMZN-puvvadar,

I'll take a look into this shortly, it's been a while so I need a refresher! Re the second question, from memory, I believe that would be accurate though unlikely and comparatively low risk compared to the initial problem of data corruption.

AMZN-puvvadar:
Hey @tkgdhughes,

We've decided we're not going to take this as is. It's certainly an improvement but we want to get to 100% reliability here. We'll continue working on the issue with this as reference. Thanks!

Leaving this open if you're able to let us know on the first point from the earlier questions as it is still a curiosity.

PR #88, opened 2018-03-26
36) Fix for detecting application focus issues

tkgdhughes:
### Description

Mouse and Keyboard focus detection fixes.

AZ input system uses WinApi ::GetFocus() to detect if the application has focus, but that function only tells which "control" has the keyboard focus. It is possible for an application to be in the background, not receiving input, but the ::GetFocus() call still returning a valid window handle (meaning - when the application becomes active, this control will receive input), which can be easily replicated.

To fix this, an extended version of::GetFocus() WinApi call, GetFocusEx(), was added.

### Detail
::GetFocus() is used to detect if application is active (in the foreground).

What ::GetFocus() really does though is return the window with the KEYBOARD focus for the CURRENT THREAD's message queue.
This basically means that the call returns the current control within the application window that will receive the keyboard events.
Usually, when the main application window is sent to background, the text focus is reset (::GetFocus() returns nullptr), but not always - it is possible for the application to be in the background, yet ::GetFocus() to continue returning a valid window handle.

The MSDN documentation is somewhat brief and vague in this area, making it easy to use this call incorrectly. It does recommend using ::GetForegroundWindow() instead to get the window with which the user is currently working.

GetFocusEx() uses ::GetForegroundWindow() to verify if the handle from::GetFocus() should be used - if ::GetFocus() handle is valid, but the application is not active, ignore it.

### DIFF
--- a/dev/Code/CryEngine/CryCommon/CrySystemBus.h
+++ b/dev/Code/CryEngine/CryCommon/CrySystemBus.h
@@ -60,6 +60,7 @@ class CrySystemRequests
 public:
     //! Get CrySystem
     virtual ISystem* GetCrySystem() = 0;
+    virtual void* GetMainWindowHandle() = 0;
 };
 using CrySystemRequestBus = AZ::EBus<CrySystemRequests>;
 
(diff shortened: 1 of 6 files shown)


gFunc:
Hi @tkgdhughes, I'm having difficulty repro'ing this problem in 1.13 and latest.
Are there more reliable steps you can share?  Thanks!

gFunc:
Thanks!  I'll open up a few clients and try alt-tabbing around :)

gFunc:
Hmm, I've tried opening and alt+tabbing around 3 different clients, but wasn't able to witness any strange behavior.  Let me know if there's any extra information available for a repro. 
Thanks!
Gene

PR #89, opened 2018-03-27
40) Add debug rendering modes to display missing LODs and objects with too few LODs

tkgdhughes:
### Description 

- Added debug render modes 24 and 25. These will highlight models which have no LODs assigned, and models which should be on a lower LOD but don't have one assigned.
- Added a new e_DebugDrawLodMinTriangles cvar which is used to selectively ignore small objects in the new debug rendering modes.
- Fixed debug render mode 3 which is meant to flash objects with a single LOD.

This change helped us to optimize our scene by easily finding models which had incorrect or non-existent LODs. The debug output for shadows also proved very useful as we found they were a consistent source of problems for us.


AMZN-alexpete:
Hey @tkgdhughes,

Thank you for the feature. We will evaluate the code and get back to you soon.

PR #90, opened 2018-03-27
39) Fix for crash due to slice instantiation

tkgdhughes:
### Description

If slices are instantiated from an OnAssetReady callback, the slice instantiation queue can be modified whilst iterating causing a crash.
The fix presented is to temporarily swap the queue for a blank one during the processing, so that new slices do not invalidate the iterators, and then recombine the queues after iterating.


ranilianaws:
Hello @tkgdhughes ,

We already have this fixed in a future version of Lumberyard. We can not take this as is unfortunately. Thank you for the contribution.

PR #91, opened 2018-03-27
m_queuedSliceInstantiations container can be changed while iterating

ximura:
In case another slice is added to m_queuedSliceInstantiations from another thread, iterator of m_queuedSliceInstantiations would be invalidated in instantiateCallback lamba.

So we use copy of m_queuedSliceInstantiations in lambda, in this way tick thread doesn't affect thread that call EntityContext::InstantiateSlice


ranilianaws:
Hello @ximura ,

We have already fixed this issue in a future version of Lumberyard. Thank you for your contribution!

PR #93, opened 2018-03-27
18) Improve pending component logging

tkgdhughes:
### Description

This is a very simple one but one that saved a lot of time debugging component related issues; simply add the name of the component to the pending component composition warning.


AMZN-puvvadar:
Hey Dan,

Thanks for the submission! We've actually already gone a different direction with that particular block of code internally so we won't be able to accept this one.

PR #94, opened 2018-03-27
30) Fix for GridMate crash

tkgdhughes:
### Description 

Fix for crash in GridMember::OnKick which can occur if a kick message is received but the host has already closed the session. Fix is to null check the session and session->GetHost() calls to avoid unsafe access. 
Issue can be reproduced by sending 2 Kick message in quick succession, also may occur if the client voluntarily leaves the session at the same time as the host trying to kick them.


ranilianaws:
Hey @tkgdhughes ,

Thank you for the contribution! We will evaluate the code and get back to you as soon as possible.

ranilianaws:
Hey @tkgdhughes ,

Thank you for the contribution. We accepted the code change into a future release of LY.

PR #95, opened 2018-03-28
1) Fix for D3D exception

tkgdhughes:
### Description 

Added nullptr protection in d3d11 deferred texture streaming. This became visible in our project due to enabling constant texture streaming in the level.

The diff looks a bit funny for me but all that has been added is:

```c++
if (tp->m_pFileTexMips->m_pPoolItem)    ///< Don't upload if the source is nullptr it'll just cause an exception
{
    ...
}
```


ranilianaws:
Hey @tkgdhughes ,

ranilianaws:
Hey @tkgdhughes ,

Thank you for the contribution. We have accepted the code as is and it will be included in a future release of Lumberyard.

PR #96, opened 2018-03-28
15) Add pump window message if starved

tkgdhughes:
### Description 

Add warning message "PumpWindowMessage not called for %.2fms" when the message queue is not pumped frequently enough. This is a bad situation so a warning should be emitted.


AMZN-alexpete:
Hey @tkgdhughes,

AMZN-alexpete:
We've decided not to take this enhancement at this time because we haven't identified a repro for this and also need to add tests.  Thank you for submitting this feature!

PR #97, opened 2018-03-29
39) Fix for unproject not working

tkgdhughes:
### Description 
Fix for un-projection not working when character capsules are overlapping, resulting in creation failure.

### Test
This is fairly easy to see working. 

**Before the fix:**
I would suggest adding some logging like the commented out logs provided in this change, it makes it easy to see when unprojection starts, ends, and fails.

- Create a new level in the Starter Game
- Add in two versions of the PlayerSlice
- Make sure they are overlapping
- Setup unprojection on them, I set it to about 2.0m and the direction to {0, 0, 0} to allow auto-unproject to work
- Run the game, note that Jack doesn't move, and if you added logs you will see something like:
```
(Unproj) - Unprojecting start
(Unproj) - Unprojecting fail
[Warning] (CharacterPhysicsComponent) - Setting dimensions failed.  The character may not have enough room.
```

**After the fix:**
- Setup the level as before
- Run the game
- Note that Jack moves across and the logs look something like: 

```
(Unproj) - Unprojecting start
(Unproj) - Unprojecting res 30, step 0.578779, unproj 0.584567
(Unproj) - Unprojecting res 29, step 0.436646, unproj 1.025579
(Unproj) - Unprojecting res 28, step 0.000000, unproj 1.025579
(Unproj) - Unprojecting done
```

### DIFF
--- a/dev/Code/CryEngine/CryPhysics/livingentity.cpp
+++ b/dev/Code/CryEngine/CryPhysics/livingentity.cpp
@@ -340,10 +340,21 @@ got_unproj:
             else if (pentlist[i]->m_parts[0].flags & collider_flags && !IgnoreCollision(m_collisionClass, pentlist[i]->m_collisionClass))
             {
                 CLivingEntity* pent = (CLivingEntity*)pentlist[i];
-                if (fabs_tpl((pos.z + hCollider - hPivot) - (pent->m_pos.z + pent->m_hCyl - pent->m_hPivot)) < newdim.z + pent->m_size.z &&
-                    len2(vector2df(pos) - vector2df(pent->m_pos)) < sqr(newdim.x + pent->m_size.x))
+                float myColliderHalfHeight = newdim.z + (newdim.x * bCapsule);
+                float theirColliderHalfHeight = pent->m_size.z + (pent->m_size.x * pent->m_bUseCapsule);
+                float verticalCenterOfMyCollider = pos.z + hCollider - hPivot;
+                float verticalCenterOfTheirCollider = pent->m_pos.z + pent->m_hCyl - pent->m_hPivot;
+                float distanceBetweenVerticalCentersOfColliders = fabs_tpl(verticalCenterOfMyCollider - verticalCenterOfTheirCollider);
+                Vec2 planarCenterToCenterOfColliders = vector2df(pent->m_pos) - vector2df(pos);
+                float squareDistanceBetweenPlanarCentersOfColliders = len2(planarCenterToCenterOfColliders);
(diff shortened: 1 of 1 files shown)


dev/Code/CryEngine/CryPhysics/livingentity.cpp
```diff
@@ -340,10 +340,21 @@ float CLivingEntity::UnprojectionNeeded(const Vec3& pos, const quaternionf& qrot
...
+                float verticalCenterOfTheirCollider = pent->m_pos.z + pent->m_hCyl - pent->m_hPivot;
+                float distanceBetweenVerticalCentersOfColliders = fabs_tpl(verticalCenterOfMyCollider - verticalCenterOfTheirCollider);
+                Vec2 planarCenterToCenterOfColliders = vector2df(pent->m_pos) - vector2df(pos);
+                float squareDistanceBetweenPlanarCentersOfColliders = len2(planarCenterToCenterOfColliders);
+                if (distanceBetweenVerticalCentersOfColliders < myColliderHalfHeight + theirColliderHalfHeight &&
+                    squareDistanceBetweenPlanarCentersOfColliders < sqr(newdim.x + pent->m_size.x))
                 {
-                    return 1E10;
+                    // Pick the best direction to unproject in the XY plane
+                    Vec2 safeUnprojDir = Vec2(cry_random(-1.0f, 1.0f), cry_random(-1.0f, 1.0f)).GetNormalizedSafe(Vec2(0.0f, 1.0f));
```
ranilianaws: Can this be simplified to a 1,0 vec2 instead? Is randomization required here?

tkgdhughes: Morning @ranilianaws , 

The rand is required because this code is entered when it was otherwise impossible to work out which direction to unproject, e.g. when 2 objects occupy exactly the same position. Using a fixed value here would mean we are assuming that it is safe to unproject in that direction. If it fails then it would just keep retrying in the same direction until the retry count expired then it would bail out. By using a random direction instead, every retry will try a different direction until it finds one that works. An example use case is when spawning 50 AI characters on the same spot. With the randomisation, they end up spawning in a circle around the spot, but with a fixed direction you would end up with a long line of characters until there was no more room. The random version ensures that all 50 characters are reasonably close to the desired spawn point.

ranilianaws: Great, thank you!

ranilianaws:
Hey @tkgdhughes ,

We have accepted the change into a future release of LY. Thanks for the contribution.

PR #98, opened 2018-03-29
41) Removed arbitrary limit on texture streaming pool. 

tkgdhughes:
### Description 

Removed arbitrary limit on texture streaming pool.  Allows full allocation (if set) of GPU RAM, but will assert and warn when going over physical and proportional limits.


AMZN-alexpete:
Hi Dan, 
Thank you for this code change, it will be included in an upcoming release of Lumberyard.
Cheers!

PR #100, opened 2018-03-29
43) Fix for render thread crash

tkgdhughes:
### Description 

Fix for a crash we were seeing at the end of rounds. 

Make the render thread lock the mutex inside ::relinktail when the list being modified; it is a fill list not a process list as these are always accessed across multiple-threads.


AMZN-alexpete:
Hi Dan, thank you for this fix, it will be included in an upcoming version of Lumberyard!

PR #101, opened 2018-03-29
44) Fix text component memory leak

tkgdhughes:
### Description 

Fix for the UiTextComponent leaking memory due to font mismatching.
The default font is registered as 'default-ui':
```
    static const AZStd::string DefaultUi("default-ui");
    m_fontFilename.SetAssetPath(DefaultUi.c_str());
```
But some fonts come from the asset system here with 'fonts/' prefixed. This causes a mismatch in Init and a memory leak. 

The solution is slightly hacky, but this special default case is now handled.

### DIFF
--- a/dev/Gems/LyShine/Code/Source/UiTextComponent.cpp
+++ b/dev/Gems/LyShine/Code/Source/UiTextComponent.cpp
@@ -1896,12 +1896,19 @@ void UiTextComponent::Init()
         return;
     }
 
-    // if the font is not the one specified by the path (e.g. after loading using serialization)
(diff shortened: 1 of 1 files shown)


ranilianaws:
Hey @tkgdhughes ,

tkgdhughes:
Morning @ranilianaws! 

I've just been stepping through this line by line and it looks like I missed some changes which were made in the 1.12 update to CryFont.cpp which fix this:

```
if (!AddFontFamilyToMaps(pFontFamilyName, fontFamily->familyName, fontFamily))
{	             
+                SAFE_RELEASE(pFont);
                 return nullptr;	                     
}
```

Thanks for the review, I'll close this one.
Dan

PR #102, opened 2018-04-03
45) Fix for level unload crashes

tkgdhughes:
### Description

We were seeing crashes during level unloading when a particle effect reference in Lua was triggered immediately before a level unload.

The level unloading process would destroy particle emitters which were still referenced in Lua and waiting to be garbage collected. When the garbage collector is subsequently triggered, the particle emitter smart pointers will be pointing to invalid memory. Triggering an extra GarbageCollect ensures that the emitters referenced in Lua are cleaned up appropriately before being destroyed.


ranilianaws:
Hey @tkgdhughes ,

We have accepted the code into a future release of LY. Thank you for the contribution.

PR #103, opened 2018-04-03
46) Add particle physics component, 47) Add ability to disable collision with player

tkgdhughes:
This is a bit of a two in one push because the two changes implement some of the same interface so I thought I would bundle them to make it a bit easier to integrate.

### 46) Description 
Added a new particle physics component. 
This implements a physics entity of type PE_PARTICLE, which according to the old Cryengine documentation is well suited for small fast moving objects, for example, projectiles. It is lightweight, and uses a ray sweep for collision detection, meaning that it cannot teleport through walls and characters when moving at high speed.

### 47) Description
Added an option to disable collisions between character capsules and RigidPhysicsComponents. Our use case for this was such that a grenade can miss the character closely, for instance through the legs, and that it can be thrown from a closer distance.

### DIFF
--- a/dev/Code/Framework/AzFramework/AzFramework/Physics/PhysicsComponentBus.h
+++ b/dev/Code/Framework/AzFramework/AzFramework/Physics/PhysicsComponentBus.h
@@ -301,6 +301,9 @@ namespace AzFramework
         //! Indicates whether this component can interact with proximity triggers
         bool m_interactsWithTriggers = true;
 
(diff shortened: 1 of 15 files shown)


ranilianaws:
Hello @tkgdhughes ,

ranilianaws:
Hey @tkgdhughes ,

We have accepted the code change into a future release of LY. Thanks for the contribution.

PR #104, opened 2018-04-03
48) Fix physics components not reporting state change

tkgdhughes:
### Description 

Fix for Component Entities not reporting physics state changes when they are updated via setting the transform. 

A rigid body component property has been added for enabling external state change monitoring, so that physics events are generated when rigid bodies are moved externally, for example, due to a direct transform modification. This property defaults to off. Setting the new property to true allows external systems, in our particular case Kythera, to listen to physics updates. In the case of Kythera, this was used to mark the navmesh as dirty and regenerate it when doors open or objects move.


tkgdhughes:
Hey @ranilianaws, any update on this?

ranilianaws:
Hey @tkgdhughes ,

Thanks for the contribution. This was accepted and will be included in a future release of LY.

PR #109, opened 2018-04-05
54) Fix for stall on Windows

tkgdhughes:
### Description

This change switches sleeps to semaphores in the CryThread_Windows impl plus adds better profiling markers.

In CryThread_windows.h the ProducerConsumerQueueBase is implemented to use semaphores rather than sleeping. The problem here was that the use of CryLowLatencySleep could take up to a whole schedule quantum (16ms) to wake up causing a stall across the entire game.

In CryThread_pthreads.h a ProducerConsumerQueueBase class has been added which is used as a base for SingleProducerSingleConsumerQueueBase and N_ProducerSingleConsumerQueueBase, this is simply a dummy base class used to match the windows interface.

### DIFF
--- a/dev/Code/CryEngine/CryCommon/CryThread.h
+++ b/dev/Code/CryEngine/CryCommon/CryThread.h
@@ -653,6 +653,8 @@ namespace CryMT
 
         m_arrBuffer = alias_cast<T*>(CryModuleMemalign(nSize * sizeof(T), 16));
         m_nBufferSize = nSize;
+
+        CryMT::detail::ProducerConsumerQueueBase::Init(nSize);  ///< Set up the queue
     }
 
     ///////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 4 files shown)


ranilianaws:
Hey @tkgdhughes,

ranilianaws:
Hello @tkgdhughes ,

We believe that this solution might not work on other platforms. Has this been tested on a pthreads platform?

tkgdhughes:
Morning @ranilianaws,

This change has been compiled into our linux dedicated server but not thoroughly tested, however, the implementation for pThreads is the same as before and does nothing so we do not think that this will be a problem. My only way to test it here in the office would be to switch to a pThread implementation on Windows and I'm not sure that is supported! What are your concerns?

Thanks

ranilianaws:
Hey @tkgdhughes ,

Thanks for the contribution. We have accepted the change into a future release of LY.

PR #110, opened 2018-04-05
55) Add pre-render step to render interface.

tkgdhughes:
### Description

Added support for a PreRender step when rendering mesh or skinned mesh components. This is useful in a variety of situations; our use case was for implementing a silhouette component which will fill out SRenderParams::nHUDSilhouettesParams in OnMeshPreRender.


ranilianaws:
Hey @tkgdhughes,

ranilianaws:
Hey @tkgdhughes ,

We made slight modifications to the code so that the EBUS is fired at the top of the Render function instead of having a PreRender function. Thanks for the contribution - it will be included in a future release of LY.

PR #111, opened 2018-04-05
57) Fix for random @devassets@ folders appearing in dev folder

tkgdhughes:
### Description 

When adding an external layer via the layers tab of the rollup bar, on saving a @devassets@ folder will appear in the root dev folder. This change correctly resolves the path to save into the correct project and level folder.

GetGameEngine()->GetLevelPath() uses an aliased path (like "@devassets@/foo/bar") instead of an absolute one. CFileUtil is a bit inconsistent, some methods work with aliased paths (i.e. CFileUtil::OverwriteFile()) some don't (CFileUtil::CreateDirectory()). Because CreateDirectory() is used in this code, we need to resolve the path first using ::GamePathToFullPath()


ranilianaws:
Hey @tkgdhughes,

ranilianaws:
Hey @tkgdhughes ,

Thanks for the contribution. We have accepted the code into a future release of LY.

PR #112, opened 2018-04-06
deque insert

ximura:
add insert with initialization list from standard
http://en.cppreference.com/w/cpp/container/deque/insert


ranilianaws:
Hello @ximura ,

We will evaluate the code and get back to you soon. Thank you for the contribution.

ranilianaws:
Hey @ximura ,
We've accepted the code change into a future version of LY. Thanks again for your contribution!

PR #115, opened 2018-04-06
59) Add a method to adjust local lag amount via CVars.

tkgdhughes:
### Description 

Added "gm_localLagAmount" CVar which can be used to set the local lag via the OnLocalLagChanged callback.


ranilianaws:
Hey @tkgdhughes ,

We have accepted the code change into a future release of LY. Thank you for the contribution!

PR #119, opened 2018-04-09
2) Add "Instant" Carrier Thread Response.

tkgdhughes:
### Description 

Added "Instant" Carrier Thread Response

By default, GridMate's messaging API (Carrier) is updated every 30ms. This change implements a method and corresponding CVar to enable an instant response to IO events which will wake up the Carrier thread immediately. 

The CVar ("gm_carrierThreadInstantResponse") when used on the Host will update and propagate to all Clients. By default this is turned off.

CVar can be used in AutoExec.cfg

With this turned off, the RTT of MP games will never be lower than 30ms (even on localhost). Turning this on will reduce the minimum RTT to ~1ms (fluctuates), however, will use more bandwidth (not currently measured).

### DIFF
--- a/dev/Code/Framework/GridMate/GridMate/Carrier/Carrier.cpp
+++ b/dev/Code/Framework/GridMate/GridMate/Carrier/Carrier.cpp
@@ -538,6 +538,8 @@ namespace GridMate
         CTM_DISCONNECT,
         CTM_DELETE_CONNECTION,
         CTM_HANDSHAKE_COMPLETE,
+        CTM_INSTANT_THREAD_RESPONSE_ENABLE,
+        CTM_INSTANT_THREAD_RESPONSE_DISABLE,
     };
 
     enum MainThreadMsg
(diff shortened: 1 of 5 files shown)


ranilianaws:
Hey @tkgdhughes,

ranilianaws:
Hey @tkgdhughes ,

I think we already have instant response enabled "by default". Look at Carrier.h's CarrierDesc structure:
m_threadInstantResponse(true).

Doesn't this satisfy this request already? We don't see a reason to change that behavior at runtime.

Let us know.

tkgdhughes:
Morning @ranilianaws,

We were changing this at runtime for the sake of our daily online tests, it was useful to be able to toggle the property and see how various online systems responded. Would it help if we marked the CVar as `VF_DEV_ONLY`?

ranilianaws:
Hey @tkgdhughes ,

The team evaluated the code and have decided not to take it due to the added complexity however we will keep the use case in mind in the future. Thanks for the contribution!

PR #120, opened 2018-04-09
22) A fix for serialized fields of type AZ::Data::Asset<T> not loading data automatically for cloned entities

tkgdhughes:
### Description 

A fix for serialized fields of type AZ::Data::Asset<T> not loading data automatically for cloned entities, even when the asset flags are set to OBJECTSTREAM_PRE_LOAD or OBJECTSTREAM_QUEUE_LOAD.

This breaks dynamic slice spawning because slice instantiation involves cloning entities of an archetype slice. Similar code existed in LY1.10 in AssetEventHandler::OnWriteEnd(), but that class was removed in LY1.12. Without this code, components of slice entities that reference assets do not work correctly, because asset data is never set. An alternative workaround for this issue is to call asset.QueueLoad() from every component that uses assets (despite the asset flags indicating that data will be loaded automatically).

### DIFF
--- a/dev/Code/Framework/AzCore/AzCore/Asset/AssetSerializer.cpp
+++ b/dev/Code/Framework/AzCore/AzCore/Asset/AssetSerializer.cpp
@@ -102,6 +102,31 @@ namespace AZ {
         asset->m_assetHint = assetHint;
         asset->UpgradeAssetInfo();
    
+        /*
+        * A fix for serialized fields of type AZ::Data::Asset<T> not loading data automatically for cloned entities,
+        * even when the asset flags are set to OBJECTSTREAM_PRE_LOAD or OBJECTSTREAM_QUEUE_LOAD.
+        * This breaks dynamic slice spawning, because slice instantiation involves cloning entities of an archetype slice.
+        * Similar code existed in LY1.10 in AssetEventHandler::OnWriteEnd(), but that class was removed in LY1.12.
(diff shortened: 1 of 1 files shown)


ranilianaws:
Hey @tkgdhughes,

ranilianaws:
Hmm, by breaks dynamic slice spawning - can you be more specific in terms of what breaks? We will need to evaluate this one further as the area of code is quite risky.

tkgdhughes:
Hey @ranilianaws, I forgot about this today, I'll get you an expanded description on Monday!

tkgdhughes:
Hey @ranilianaws,

We're just putting together a test scene for this. It's super easy to repro in DRG so we're trying to distill that down.

tkgdhughes:
Morning @ranilianaws, The plot thickens... my colleague found what looks to be this fix in [the 1.12 release](https://github.com/aws/lumberyard/commit/164512f8d415d6bdf37e195af319ffe5f96a8f0b#diff-5c6f3235dbaa642b078066b953db1350) but for some reason, we are still hitting the same problem in our game, but not in Vanilla LY. I'm going to investigate further this end and see what I can come up with. I'll message back here with more details.

tkgdhughes:
Hi @ranilianaws,

I'm going to close this PR and submit a new one. I'm struggling to get to the root of it. If I do find the issue I will re-open and let you guys know!

PR #121, opened 2018-04-09
60) Enabling RENDERER_ENABLE_BREAK_ON_ERROR caused compile errors

tkgdhughes:
### Description

- Enabling RENDERER_ENABLE_BREAK_ON_ERROR caused compile errors for DX12, OpenGL.
- Case statements in ToString were using non-literals, so illegal syntax.
- Format changed to enable double-clicking an error to jump to the source

### DIFF
--- a/dev/Code/CryEngine/RenderDll/XRenderD3D9/DriverD3D.cpp
+++ b/dev/Code/CryEngine/RenderDll/XRenderD3D9/DriverD3D.cpp
@@ -62,88 +62,81 @@
 #include <windows.h>
 #endif
 #if RENDERER_ENABLE_BREAK_ON_ERROR
+#include <d3d9.h>
 namespace detail
 {
     const char* ToString(long const hr)
     {
-        switch (hr)
-        {
-        case D3DOK_NOAUTOGEN:
+        if (D3DOK_NOAUTOGEN == hr)
             return "D3DOK_NOAUTOGEN This is a success code. However, the autogeneration of mipmaps is not supported for this format. This means that resource creation will succeed but the mipmap levels will not be automatically generated";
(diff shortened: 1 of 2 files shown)


dev/Code/CryEngine/RenderDll/XRenderD3D9/DriverD3D.cpp
```diff
@@ -155,7 +148,7 @@ namespace detail
         {
             return true;
         }
-        CryLogAlways("[%s:%d] d3d error: '%s'", file, line, ToString(hr));
+        CryLogAlways("%s(%d): d3d error: '%s'", file, line, ToString(hr));
```
tkgdhughes: Format changed to enable double-clicking an error to jump to the source

dev/Code/CryEngine/RenderDll/XRenderD3D9/DriverD3D.cpp
```diff
@@ -62,88 +62,81 @@
 #include <windows.h>
 #endif
 #if RENDERER_ENABLE_BREAK_ON_ERROR
+#include <d3d9.h>
```
tkgdhughes: D3DERR_XXX not defined

ranilianaws:
Hey @tkgdhughes,

tkggwatson:
@ranilianaws

Case statements in ToString were using non-literals, so illegal syntax.

In dx12, the values are not literals, so cannot be used in case statements. The new version works with both dx11 and dx12.

ranilianaws:
Hey @tkgdhughes ,

We have accepted the code change and it will be included in a future release of LY. Thank you for the contribution!

PR #122, opened 2018-04-09
63) Bug fix for physics entities failing to create.

tkgdhughes:
### Description 

Bug fix for physics entities failing to create due to their initial position being set on a thread rather than at creation time. In our case, this was resulting in floating rockets and characters that could not move.

- At creation, specify 0x5AFE as the foreign data, which causes the Cry physics system to set the initial parameters immediately rather than putting them on to the job queue.
- Added the pe_params_foreign_data SetParams call after creation, to put the component-entity data into the foreign data (by this point, the 0x5AFE has been used and is no longer needed).


tkgdhughes:
Hey @ranilianaws, any update on this?

AMZN-alexpete:
Hey @tkgdhughes,

PR #124, opened 2018-04-10
66) Fix for the scan lines post process GFX using "texture not found"

tkgdhughes:
### Description 

A fix for the scan lines post-process GFX using "texture not found" placeholder texture in Release.

VisualArtifacts post process FX does not use any mask texture by default. This is a valid, handled case, yet the effect was using "Texture not found" texture in release builds.


ranilianaws:
Hey @tkgdhughes ,

AMZN-alexpete:
This fix is available in version 1.24, thanks again!

PR #127, opened 2018-04-10
69) Flag to disable streaming on a material

tkgdhughes:
### Description 

Added a MTL_FLAG_DO_NOT_STREAM flag which, post material load, will disable texture streaming if set on the material or a sub-material.

### DIFF
--- a/dev/Code/CryEngine/Cry3DEngine/MatMan.cpp
+++ b/dev/Code/CryEngine/Cry3DEngine/MatMan.cpp
@@ -427,6 +427,24 @@ _smart_ptr<IMaterial> CMatMan::LoadMaterialInternal(const char* sMtlName, bool b
     {
         pMtl = CreateMaterialPlaceholder(name.c_str(), nLoadingFlags, szDefaultSolid);
     }
+    
+    if (pMtl)
+    {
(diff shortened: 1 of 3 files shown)


ranilianaws:
Hey @tkgdhughes ,

Can you elaborate on the specific use case that utilizes this feature? How are you using it in your game? We are trying to evaluate the usefulness of this code as it could potentially lead to frame spikes by loading the full resolution textures instead of streaming in the lowest mips first.

tkgdhughes:
Hiya @ranilianaws,

This was used in DRG to force a few important materials to pre-load and ensure we never had any issues with them popping in during a game.

- We placed a `force_load_materials.slice` into our map which had references to materials which we had set the `MTL_FLAG_DO_NOT_STREAM` flag on. 
- Our character materials and some of our large environment materials were placed on this slice. 
- Prior to this change, we were frequently getting issues where the round would start before the materials were loaded in which meant the character materials would pop in at any time from 1-2 seconds to 30 seconds _(the most we saw)_ after the round started.

ranilianaws:
Hey @tkgdhughes ,

Thanks for the contribution and for the clarification. We have accepted the code into a future release of LY.

PR #128, opened 2018-04-12
Fix file name display in error message

alex-leleka:
Error message was showing %s instead of file name.
http://doc.qt.io/qt-5/qobject.html#tr


ranilianaws:
Hey @alex-leleka ,

PR #129, opened 2018-04-12
64) Fix for incorrect time interval being used in physics

tkgdhughes:
### Description

Fix for the incorrect time interval being used to process entities.

`time_interval` is being used and modified by rigid bodies before being passed to living/independent entities for processing. This results in an incorrect time_interval being used causing incorrect motion in certain situations. This is fixed by passing the unmodified `time_interval_org` for processing. 

In our use case, we model projectiles using CryEngine particle physics. This bug resulted in projectiles having an incorrectly variable speed.


AMZN-alexpete:
Hey @tkgdhughes,

PR #131, opened 2018-04-13
max_element

ximura:
add max_element with predicat from std
http://en.cppreference.com/w/cpp/algorithm/max_element


ranilianaws:
Hey @ximura ,

ranilianaws:
Hey @ximura ,

PR #133, opened 2018-04-16
73) Fix for missing callstacks on memory leak tracking.

tkgdhughes:
### Description 

This is a simple change but one which will save another developer a bit of time finding this vector.
- Increased the debug memory usage as we were overflowing the module info vector. This increases memory usage in debug by about 400KB.


AMZN-alexpete:
Hi @tkgdhughes,  Thank you for sending us this fix.  There was a separate issue that was causing the overflow.  The bug was that it was not reusing module slots correctly and would always eventually overflow.  This fix will be in an upcoming release of Lumberyard.

PR #134, opened 2018-04-16
74) Fix for GridMate carrier crash

tkgdhughes:
### Description

Fix for GridMate carrier crash caused by the use of an invalid variable in Release. The change is to not just assert and carry on, but rather to return Disconnected as if an Invalid id was passed in.


AMZN-alexpete:
Hey @tkgdhughes,

AMZN-alexpete:
Closing this request as it is fixed, but in a different manner:
[https://github.com/aws/lumberyard/blob/master/dev/Code/Framework/GridMate/GridMate/Carrier/Carrier.cpp#L4582]

PR #135, opened 2018-04-16
75) Add explicit support for marshalling the "size_t" type with GridMate.

tkgdhughes:
### Description 

Add explicit support for marshalling the `size_t` type with GridMate.
On Windows, this isn't needed as `size_t` is `unsigned long long` which is already covered by having  `AZ::u64` in the list of fundamental types for marshalling.
On Linux `size_t` is `unsigned long`, which is also 8 bytes, but not covered by `AZ::u64` so needs to be explicitly added.

### DIFF
--- a/dev/Code/Framework/GridMate/GridMate/Serialize/DataMarshal.h
+++ b/dev/Code/Framework/GridMate/GridMate/Serialize/DataMarshal.h
@@ -65,7 +65,8 @@ namespace GridMate
                 AZStd::is_same<Type, AZ::s8>::value  ||
                 AZStd::is_same<Type, AZ::s16>::value ||
                 AZStd::is_same<Type, AZ::s32>::value ||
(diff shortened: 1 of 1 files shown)


ranilianaws:
Hey @tkgdhughes,

ranilianaws:
Hey @tkgdhughes ,

Thanks for the contribution but we will not accept the code as is as since size_t isn't a standardized explicitly sized type across platforms, this marshaler would create a situation where a user could unknowingly add a size_t type to marshal on a platform which uses unsigned long on a client and unsigned long long on the server. By not supporting a marshaler for size_t it prevents a user from making an easy to make mistake and forces them to either implement their own marshaler OR to change the data type to a more explicit native type that would be the same on different platforms.

tkgdhughes:
Fair enough, thanks for the feedback.

PR #137, opened 2018-04-18
79) Add ignored entities to trigger area component

tkgdhughes:
### Description 

Similarly to how TriggerAreaComponents can ignore entities based on tag, functionality has been added to ignore entities based on entityId. This was used extensively from bespoke Lua components for trigger area management in DRG.

### DIFF
--- a/dev/Gems/LmbrCentral/Code/Source/Scripting/EditorTriggerAreaComponent.cpp
+++ b/dev/Gems/LmbrCentral/Code/Source/Scripting/EditorTriggerAreaComponent.cpp
@@ -46,10 +46,15 @@ namespace LmbrCentral
                     ->DataElement(AZ::Edit::UIHandlers::ComboBox, &TriggerAreaComponent::m_activationEntityType, "Activated by", "The types of entities capable of interacting with the area trigger.")
                         ->EnumAttribute(TriggerAreaComponent::ActivationEntityType::AllEntities, "All entities")
                         ->EnumAttribute(TriggerAreaComponent::ActivationEntityType::SpecificEntities, "Specific entities")
+						->EnumAttribute(TriggerAreaComponent::ActivationEntityType::IgnoreEntities, "Ignore entities")
                         ->Attribute(AZ::Edit::Attributes::ChangeNotify, &TriggerAreaComponent::OnActivatedByComboBoxChanged)
                     ->DataElement(AZ::Edit::UIHandlers::Default, &TriggerAreaComponent::m_specificInteractEntities, "Specific entities", "List of entities that can interact with the trigger.")
                         ->Attribute(AZ::Edit::Attributes::Visibility, &TriggerAreaComponent::UseSpecificEntityList)
-                        ->Attribute(AZ::Edit::Attributes::AutoExpand, true);
+                        ->Attribute(AZ::Edit::Attributes::AutoExpand, true)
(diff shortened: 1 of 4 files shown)


ranilianaws:
Hey @tkgdhughes,

dev/Gems/LmbrCentral/Code/Source/Scripting/EditorTriggerAreaComponent.cpp
```diff
@@ -46,10 +46,15 @@ namespace LmbrCentral
...
                         ->EnumAttribute(TriggerAreaComponent::ActivationEntityType::AllEntities, "All entities")
                         ->EnumAttribute(TriggerAreaComponent::ActivationEntityType::SpecificEntities, "Specific entities")
+						->EnumAttribute(TriggerAreaComponent::ActivationEntityType::IgnoreEntities, "Ignore entities")
                         ->Attribute(AZ::Edit::Attributes::ChangeNotify, &TriggerAreaComponent::OnActivatedByComboBoxChanged)
                     ->DataElement(AZ::Edit::UIHandlers::Default, &TriggerAreaComponent::m_specificInteractEntities, "Specific entities", "List of entities that can interact with the trigger.")
                         ->Attribute(AZ::Edit::Attributes::Visibility, &TriggerAreaComponent::UseSpecificEntityList)
-                        ->Attribute(AZ::Edit::Attributes::AutoExpand, true);
+                        ->Attribute(AZ::Edit::Attributes::AutoExpand, true)
+					->DataElement(AZ::Edit::UIHandlers::Default, &TriggerAreaComponent::m_ignoredEntities, "Ignore entities", "List of entities which will not interact with the trigger.")
```
ranilianaws: This breaks serialization as adding a field without upping the version number will cause the serializer to attempt to load the IgnoreInteractEntities field in Version 2 TriggerAreaComponents

dev/Gems/LmbrCentral/Code/Source/Scripting/TriggerAreaComponent.cpp
```diff
@@ -353,6 +356,30 @@ namespace LmbrCentral
         }
     }
 
+	void TriggerAreaComponent::AddIgnoredEntity(const AZ::EntityId& ignoredEntityId)
+	{
+		const auto& ignoredEntityIter = AZStd::find(m_ignoredEntities.begin(), m_ignoredEntities.end(), ignoredEntityId);
```
ranilianaws: This returns an iterator type by value, so it should just be auto.

Also why not make m_ignoredEntities an unordered_set?

dev/Gems/LmbrCentral/Code/Source/Scripting/EditorTriggerAreaComponent.cpp
tkgdhughes: I'll bump that now.

dev/Gems/LmbrCentral/Code/Source/Scripting/TriggerAreaComponent.cpp
tkgdhughes: m_ignoredEntities is just following in the footsteps of the existing code for example `AZStd::vector<AZ::EntityId>     m_entitiesInsideExcludedByTags;` and `AZStd::vector<AZ::EntityId>     m_entitiesInside;`

I considered using a typedef and changing all of the existing containers to something else but reasoned that I would:
a) Make the smallest change possible.
b) Not change existing already tested code.
I know it isn't a scary change but it wasn't necessary here.

PR #138, opened 2018-04-18
78) Fix for lmbr.exe populate-appdescriptors command generating corrupt files. 

tkgdhughes:
### Description 

Fix for lmbr.exe `populate-appdescriptors` command generating corrupt files.

Previously `FlavorXml::PopulateModules` would clone the descriptor node from the in-memory v3 object stream to the on-disc v-anything XML. This is not a valid thing to do, as it would leave the SystemEntity and root node on the existing ObjectStream version, leaving us with a v2 XML file that contains a v3 descriptor. Instead, we deserialise and reserialise both the descriptor AND the system entity, and write out a whole new XML file in the latest ObjectStream version.

This has the disadvantage that it will destroy any custom system components that have been added to the SystemEntity, because lmbr.exe does not load the gem dlls and register the system components with the serialiser, so they get stripped on load as they are not recognised.

Until we come up with a better fix, this is preferable to writing out corrupt files, and system components will need to be re-added from the ProjectConfigurator after running a `populate-appdescriptors` step.

In `ObjectStream.cpp`, there are fixes for un-initialised variables overwriting the element type if no specializationTypeId attribute exists.

**Note:** 
This GitHub Pull Request _only_ has the changes for `ObjectStream.cpp`. 
The change for Flavor.cpp has been shared separately to this request.
We do not currently have the 1.13 version of Flavor.cpp as it is not distributed through GitHub so this change is against version 1.12. I have left inline comments in `Flavor.cpp` to help follow the reasoning.

### DIFF
--- a/dev/Code/Framework/AzCore/AzCore/Serialization/ObjectStream.cpp
+++ b/dev/Code/Framework/AzCore/AzCore/Serialization/ObjectStream.cpp
@@ -893,7 +893,7 @@ namespace AZ
                 }
                 m_xmlNode = next;
 
-                Uuid specializedId;
+                Uuid specializedId = Uuid::CreateNull();
                 // now parse the node
                 rapidxml::xml_attribute<char>* attr = m_xmlNode->first_attribute();
                 while (attr)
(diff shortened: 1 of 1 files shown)


ranilianaws:
Hey @tkgdhughes,

ranilianaws:
Hello @tkgdhughes ,

Thank you for the contribution. We have created an internal ticket to address the underlying issue you are experiencing and as such will not be accepting the code as is.

PR #139, opened 2018-04-19
77) Debug camera additions.

tkgdhughes:
### Description 

DebugCamera modifications, fixes and improvements. 

Note: This changes works with the legacy input and debug camera system, ie the CryLegacy Gem is enabled. The debug free-fly camera has not been updated to be controlled with the new input system (same as current 1.13 behaviour).

### Changes
- Do not send inputs to game entities if debug camera is enabled in free-fly mode. This allows for positioning of a debug camera without the player character moving around.
- Mouse wheel scales the linear speed only (and not angular speed). Typically rotational speed is not an issue whereas translation speed is. 
- Press left ctrl to slow down the camera.
- Additions to get debug camera position & orientation (we used this for teleporting the player to the debug camera position).
- Changed key bindings to use slash instead of backslash which doesn't appear to work
- Hook into console event binding to allow handling of events before the game's main loop.
- Use the UI update time so that the debug camera can still be controlled if the game is paused.
- CTRL-Slash toggles debug camera on/off (instead of on/fixed/off).
- CTRL-LSHIFT-Slash toggles fixed camera on/off, this will work when debug camera is off or in free-fly mode
-- The two changes above combine to make the functionality of the debug camera feel more consistent. Pressing CTRL-/ always turns free fly on/off, pressing CTRL-LSHIFT-/ always turns fixed mode on/off

### DIFF
--- a/dev/Code/CryEngine/CryAction/DebugCamera/DebugCamera.cpp
+++ b/dev/Code/CryEngine/CryAction/DebugCamera/DebugCamera.cpp
@@ -36,15 +36,26 @@ void ToggleDebugCamera(IConsoleCmdArgs* pArgs)
     DebugCamera* debugCamera = CCryAction::GetCryAction()->GetDebugCamera();
     if (debugCamera)
     {
-        if (!debugCamera->IsEnabled())
+        if (!debugCamera->IsEnabled() || debugCamera->IsFixed()	)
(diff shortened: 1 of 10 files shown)


ranilianaws:
Hey @tkgdhughes,

PR #140, opened 2018-04-20
80) Fix for crash due to missing physics foreign data

tkgdhughes:
### Description

Fix for a crash seen in BakeScaleIntoGeometry.

A better fix for geometry being stored as unscaled geom in physics foreign data. The code which earlies out in utils.cpp BakeScaleIntoGeometry is just a workaround for this issue and could potentially be removed now:
``` c++
    if (!pgeom->pGeom)
    {
        return 0;
    }
```

Add reference counting to the foreign physics data to ensure that unscaled geometry is not released too early. 
The fix is not quite as robust as it should be; it only works for unscaled geometry and the lifetime is only valid if the foreign data of the scaled geometry is not modified after creation. However, the data is not designed for this problem so I've added an assert to catch it and it doesn't fire.

### DIFF
--- a/dev/Code/CryEngine/CryPhysics/geoman.cpp
+++ b/dev/Code/CryEngine/CryPhysics/geoman.cpp
@@ -183,7 +183,13 @@ int CGeomManager::AddRefGeometry(phys_geometry* pgeom)
 
 int CGeomManager::UnregisterGeometry(phys_geometry* pgeom)
 {
-    WriteLock lock(m_lockGeoman);
+	WriteLock lock(m_lockGeoman);
+	return UnregisterGeometryLocked(pgeom);
+}
+
+// Internal function that assumes that the caller is already in the write lock, see ::UnregisterGeometry
+int CGeomManager::UnregisterGeometryLocked(phys_geometry* pgeom)      
+{
     AtomicAdd(&pgeom->nRefCount, -1);
     if (pgeom->nRefCount != 0)
     {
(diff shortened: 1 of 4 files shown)


dev/Code/CryEngine/CryPhysics/geoman.cpp
```diff
@@ -183,7 +183,13 @@ int CGeomManager::AddRefGeometry(phys_geometry* pgeom)
 
 int CGeomManager::UnregisterGeometry(phys_geometry* pgeom)
 {
-    WriteLock lock(m_lockGeoman);
+	WriteLock lock(m_lockGeoman);
+	return UnregisterGeometryLocked(pgeom);
```
tkgdhughes: Moved the functionality into a private locked function so it can be called recursively

dev/Code/CryEngine/CryPhysics/geoman.cpp
```diff
@@ -183,7 +183,13 @@ int CGeomManager::AddRefGeometry(phys_geometry* pgeom)
...
+}
+
+// Internal function that assumes that the caller is already in the write lock, see ::UnregisterGeometry
+int CGeomManager::UnregisterGeometryLocked(phys_geometry* pgeom)
```
tkgdhughes: Added internal function that assumes that the caller is already in the write lock

dev/Code/CryEngine/CryPhysics/geometry.h
```diff
@@ -348,7 +349,12 @@ class CGeometry
     virtual Vec3 GetCenter() { return Vec3(ZERO); }
     virtual PhysicsForeignData GetForeignData(int iForeignData = 0) { return iForeignData == m_iForeignData ? m_pForeignData : PhysicsForeignData(0); }
     virtual int GetiForeignData() { return m_iForeignData; }
-    virtual void SetForeignData(PhysicsForeignData pForeignData, int iForeignData) { m_pForeignData = pForeignData; m_iForeignData = iForeignData; }
+	virtual void SetForeignData(PhysicsForeignData pForeignData, int iForeignData)
+	{
+		AZ_Assert(m_iForeignData != DATA_UNSCALED_GEOM, "Overriding foreign data that is reference counted, will leak geometry");
```
tkgdhughes: Added assert to catch the ref-counted pointer being overridden as there is no good way of handling this.

PR #141, opened 2018-04-23
52) Fix for particles rotating horribly when they die on collision.

tkgdhughes:
### Description 

Bug fix for particles rotating horribly when they die on collision.
- Don't rotate the particle to face the normal of the hit surface.

### DIFF
--- a/dev/Code/CryEngine/Cry3DEngine/Particle.cpp
+++ b/dev/Code/CryEngine/Cry3DEngine/Particle.cpp
@@ -1515,9 +1515,6 @@ void CParticle::Update(SParticleUpdateContext const& context, float fFrameTime, 
                     stateNew.m_Loc.t = hit.pt;
 
                     fStepTime = max(fStepTime * hit.dist, context.fMinStepTime);
-
-                    // Rotate to surface normal.
-                    RotateToUpVector(stateNew.m_Loc.q, hit.n);
                     stateNew.Collide(fStepTime);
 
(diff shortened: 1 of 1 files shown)


ranilianaws:
Hey @tkgdhughes,

ranilianaws:
Hey @tkgdhughes ,

Can you give us more information on how to reproduce this? What particle settings were you using? CPU or GPU particles? Does the shape of the emitter matter?

Thanks!

tkgdhughes:
Morning @ranilianaws,

It is (thankfully) really easy to build a scene to test this before and after the fix.

- Create a new level
- Create an entity with a `Particle` component on it
- Open the particle editor
- Create a new library and a new particle effect
- Set the `Particles/Facing` attribute to be velocity so the default arrow particles will line up with the velocity nicely
- Set the `Collision/Physics` Type to `SimpleCollision`
- Set the `Collision/Collide Terrain` checkbox to be ticked
- Go back to the Entity Inspector and rotate the particle emitter component so that the particles intersect with the terrain at 45 degrees
- Note that if the `Collision/On Collide` attribute for the particle is set to `Die` the particle will rotate to the normal before dying. If it is set to `Stop` it will stop and align with the surface normal. 

When you apply our fix the particles will behave more naturally when dying or stopping.

Hope this helps!

ranilianaws:
Thanks @tkgdhughes!

We were able to reproduce the issue and accepted the code change into a future release of LY.

PR #143, opened 2018-04-25
78) Fix for endless loop of Lua script reloading bug

tkgdhughes:
### Description 

Fix for a bug where Lua scripts would endlessly reload.

It was possible, in a large project with a significant number of scripts to reload, to get into a situation where endless Lua script reloading would happen. 

**The reason for this was that:**
- The old algorithm cleared the `m_assetsAlreadyReloaded` map once all assets were requested to reload without waiting for them to be handled. 
- When there are a significant number of scripts to reload, it may happen that some of the scripts are not fully reloaded by the time `reloadFn` is finished. 
- When they were processed after `reloadFn`, their `OnAssetReloaded` callback would call `reloadFn` again, triggering a vicious circle of endless reloading.

**To fix this:**
- Instead of tracking already reloaded assets, track all reload requests and remove them from the map one by one as they are handled. 
- Once all queued reloads are handled, another full reload can be triggered.

### DIFF
--- a/dev/Code/Framework/AzCore/AzCore/Script/ScriptSystemComponent.cpp
+++ b/dev/Code/Framework/AzCore/AzCore/Script/ScriptSystemComponent.cpp
@@ -683,44 +683,63 @@ void ScriptSystemComponent::OnAssetPreReload(Data::Asset<Data::AssetData> asset)
 //=========================================================================
 // OnAssetReloaded
 // #TEMP: Remove when asset dependencies are in place
-// This function will only be called for the asset that triggered the reload (everything else is disconnected)
+// This function will only be called for the asset that triggered the reload.
+//
+// Track all reload requests and remove them from the map one by one as they are handled. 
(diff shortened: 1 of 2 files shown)


ranilianaws:
Hey @tkgdhughes,

ranilianaws:
Hey @tkgdhughes ,

Thanks for the contribution. We accepted the code and it will be included in a future version of LY.

Markersx:
@ranilianaws I'm curious about why you close accepted pull requests instead of merging. What is the benefit?

ranilianaws:
Hey @Markersx,

We close them because we don’t internally use Git and as such merging the pull request doesn’t necessarily mean that the change is included in the next release. We are aware that is quite unconventional but we are actively improving our processes. 

Thanks!

Markersx:
@ranilianaws That makes sense. Also are there plans to enable the issues tab?

ranilianaws:
Not at the moment but I will forward that feedback to the team. Thank you!

PR #144, opened 2018-04-26
44) Fixes for 3DEngineLight

tkgdhughes:
### Description

Fix for two crashes seen in 3DEngineLight:

1) Fix for referencing out of bounds light id
2) Fix thread safety of light interactions to fix collision between reset and update on level unload

### DIFF
--- a/dev/Code/CryEngine/Cry3DEngine/3DEngineLight.cpp
+++ b/dev/Code/CryEngine/Cry3DEngine/3DEngineLight.cpp
@@ -652,6 +652,8 @@ void C3DEngine::OnCasterDeleted(IShadowCaster* pCaster)
 
 void CLightVolumesMgr::Init()
 {
+	CryAutoLock< CryMutex > lock(m_dataMutex);
+
     m_bUpdateLightVolumes = false;
     for (int i = 0; i < RT_COMMAND_BUF_COUNT; ++i)
     {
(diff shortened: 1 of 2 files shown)


AMZN-alexpete:
Hi @tkgdhughes,  I've been attempting to reproduce the thread safety issue between reset and update, but both of those run from the main thread.  Is there some way you were able to trigger either function from a different thread?

AMZN-alexpete:
Is it possible the Reset() call on the main thread cleared the m_pLightVolumes that the render thread was trying to access via RT_GetLightVolumes?

tkgdhughes:
Hey @AMZN-alexpete,
I just noticed this so won't have time to look properly until tomorrow. If I remember correctly, we suspected that the reset was colliding with the render thread update during level unload. Hopefully, that may help!

AMZN-alexpete:
Hi @tkgdhughes, do you know how you were calling UnloadLevel()?  Were you using Flowgraph, custom code/EBUS, console command or something else?

tkgdhughes:
Hey @AMZN-alexpete,

We trigger some custom EBus code via a basic flow node: `EBUS_EVENT(TK::LevelManager::LevelManagerRequestBus, LoadMultiplayerLobby);` 

Which will call: `CCryAction::ExecuteCommandNextFrame` 

With the load map command as a parameter: `map MultiplayerLobby`

AMZN-alexpete:
Hi @tkgdhughes,  I've been unable to reproduce this issue or determine how it is possible for the current code to result in a race condition between Reset() and any of the functions that have added data locks or the GetLightVolumes() function which appears to be the only one called from the render thread.  My only remaining theory is perhaps there was another thread used for rendering a loading screen or something that could have triggered the race condition?

Also, it appears that the out of bounds error was submitted (and accepted) in separate pull request [https://github.com/aws/lumberyard/pull/188]  

I'm going to close this request, but please let me know if you have any more information and I will re-open.  Thank you!

PR #145, opened 2018-04-26
81) Fix for "CFlowNode::Serialize: FG 'Foo': Can't resolve value for 'Bar' <may not be severe>" warnings

tkgdhughes:
### Description

Fix for "CFlowNode::Serialize: FG 'Foo': Can't resolve value for 'Bar' <may not be severe>" warnings, which were triggered for nodes that used inputs of custom types (some of the multiplayer nodes do so). Inputs of custom types are never serialized to files (by design), so the message about them being missing during loading was superfluous.
Also added a more descriptive message to a warning in the FG loading code


tkgdhughes:
No problem, I wasn't actually sure about FG anymore. We obviously made heavy use of it and have a bunch more fixes but we pre-dated SC!

PR #146, opened 2018-05-01
Fix for game crash on r_WaterReflections=0

alex-leleka:
Game crashes if water is present on level and user activates r_WaterReflections=0. Render context is used from Main and Render threads.

### DIFF
--- a/dev/Code/CryEngine/RenderDll/XRenderD3D9/D3DFXPipeline.cpp
+++ b/dev/Code/CryEngine/RenderDll/XRenderD3D9/D3DFXPipeline.cpp
@@ -285,7 +285,6 @@ void CD3D9Renderer::EF_ClearTargetsLater(uint32 nFlags)
     //      uint8(m_pNewTarget[0]->m_pSurfDepth->pTex->GetClearColor().g));
 }
 
-
 void CD3D9Renderer::FX_ClearTargetRegion(const uint32 nAdditionalStates /* = 0*/)
 {
     assert(m_pRT->IsRenderThread());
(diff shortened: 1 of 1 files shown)


AMZN-alexpete:
Hi @alex-leleka, thank you for submitting this fix.  Do you have the call stack for this?  I believe we need to investigate why the FX_ function is being called on the main thread in the first place and look at moving the call to the render thread using an EnqueueRenderCommand call.  Also, there is an assert that should be firing earlier in that function if that function is called from the main thread.

alex-leleka:
Hello @AMZN-alexpete, i 'll provide callstack tomorrow (entry point is C3DEngine::RenderScene-> GetRenderer()->EF_EndEf3D). I also was surprised that this doesn't have assert.
You can check same function in other similar open-source engine and they have same fix working for them.

alex-leleka:
image
Debug layer message: 
D3D11 CORRUPTION: ID3D11DeviceContext::ClearRenderTargetView: Two threads were found to be executing functions associated with the same Device[Context] at the same time. This will cause corruption of memory. Appropriate thread synchronization needs to occur external to the Direct3D API (or through the ID3D10Multithread interface). 7952 and 2684 are the implicated thread ids. [ MISCELLANEOUS CORRUPTION #28: CORRUPTED_MULTITHREADING]
Reproducible: 10/10

AMZN-alexpete:
Hi @alex-leleka, we have integrated this fix and it will be included in an upcoming release of Lumberyard.  Thank you!

PR #148, opened 2018-05-08
22) Add surface id to raycast hit

tkgdhughes:
### Description 

Add the surface id to the raycast hit result.
This is a simple but often used addition which allows features such as choosing different hit impact effects per surface type. For example, using a different particle effect when shooting a steel surface as opposed to a concrete one.


ranilianaws:
Hey @tkgdhughes ,

PR #149, opened 2018-05-08
76) Fix for CRenderMesh assert due to mismatched lock/unlock

tkgdhughes:
### Description 

There appears to be a lock/unlock mismatch for threaded data in the Cloth code. This could occasionally cause an assert during the `CRenderMesh` destruction as `m_nThreadAccessCounter > 0` when quitting a map with cloth in it. The fix is to match up the lock/unlock correctly.


ranilianaws:
Hey @tkgdhughes ,

Thanks for the contribution! We will evaluate the code and get back to you soon!

ranilianaws:
Hey @tkgdhughes ,

Thanks for the contribution. We have accepted the code and it will be included in a future release of LY.

PR #150, opened 2018-05-09
Fix editor crash on sturtup in bind forward shading resources

alex-leleka:
Editor crashes on startup if BindForwardShadingResources executed before resources have been created in CreateResources. This is a rare case, but it happens sometimes.
No exact repro steps.


ranilianaws:
Hello @alex-leleka ,

Thanks for the submission. We will evaluate the code and get back to you soon.

ranilianaws:
Hey @alex-leleka ,

This has been integrated into a future release of LY. Thanks for the contribution.

PR #151, opened 2018-05-10
81) Trivial addition to allow limb IK to be enabled/disabled if necessary.

tkgdhughes:
### Description 

A trivial addition which we used for ensuring limb IK was correctly enabled/disabled in various situations. For example, when the player is dead we no longer want the IK to be aligning the feet with the ground.

### Usage
``` lua
function character_animation:OnDead(entityId, attackerEntityId)
    if(self.entityId == entityId) then
        self:RequestBaseState(EBaseState.Death)
        LimbIKComponentRequestBus.Event.EnableLimbIK(self.entityId, false)
    end
end
```


ranilianaws:
Hey @tkgdhughes ,

ranilianaws:
Hey @tkgdhughes ,

Thanks for the contribution. We evaluated the code and will include it in a future release of LY.

PR #152, opened 2018-05-10
82) Add a callback define to allow sys_affinity to be set after launch

tkgdhughes:
### Description 

A simple addition. As the title suggests we have changed the registration of the "sys_affinity" CVar to use `REGISTER_INT_CB` so that we can hook it up an `AffinityChangedCallback`.

### DIFF
--- a/dev/Code/CryEngine/CrySystem/SystemWin32.cpp
+++ b/dev/Code/CryEngine/CrySystem/SystemWin32.cpp
@@ -104,6 +104,13 @@ const char* g_szModuleGroups[][2] = {
 };
 
 //////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


ranilianaws:
Hey @tkgdhughes ,

What is the use case for this? We are trying to evaluate if we want to allow modifying the sys_affinity at runtime.

tkgdhughes:
Hey @ranilianaws,

We used this so that we could simulate lower performance clients connecting to better clients. It was pretty useful for catching some hard to reproduce bugs in the wild caused by dissimilar spec PCs. 

I update the PR with a _RELEASE define to default to the previous behaviour of no runtime modification.

PR #153, opened 2018-05-10
83) Add CloneGameEntity() request support to GameEntityContext 

tkgdhughes:
### Description

Add CloneGameEntity() request support to GameEntityContext. 
The use case for us was cloning existing consumable inventory items, such as grenades. This allowed us to clone N grenades, store them, then activate and launch them when requested by the player.

### DIFF
--- a/dev/Code/Framework/AzFramework/AzFramework/Entity/GameEntityContextBus.h
+++ b/dev/Code/Framework/AzFramework/AzFramework/Entity/GameEntityContextBus.h
@@ -92,7 +92,7 @@ namespace AzFramework
 		* Clones an existing entity in the game context.
 		* @param entity A pointer to the entity to clone.
 		* @param activate Optional parameter to skip activation.
-		* @param name
+		* @param name A name for the cloned entity.
 		*/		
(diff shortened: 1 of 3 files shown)


ranilianaws:
Hey @tkgdhughes ,

This functionality appears to be already serviced by having a dynamic slice with a SpawnerComponent.
For a inventory system case, a designer can create their grenade and other consumable items on an entity, save that entity to a slice, make it dynamic  and then just spawn that dynamic slice at runtime through the spawner component

Can the SpawnerComponent not be used for this usecase?

tkgdhughes:
Hi @ranilianaws,

This was due to how our inventory system implementation worked; it used entities instead of slices. This change gave us the flexibility to clone the entity without activating it. Not activating it was a key point as the activation would have caused physics issues due to overlapping collision meshes.

ranilianaws:
Hey @tkgdhughes ,

I see. Instead of introducing a change to the core API, you should be able to use EntityContext:CloneEntity and then call SetRuntimeActiveByDefault on the entity to achieve this functionality. Any concerns with that approach?

tkgdhughes:
At @ranilianaws,

That sounds pretty reasonable. That may have either escaped our notice first time around or maybe simply wasn't there. I'll have a bit more of a look and get back to you. Unless I come up with any good reasons I'll close this request. 

Thanks!

ranilianaws:
Closing this as a workaround was provided instead.

PR #155, opened 2018-05-14
85) Add collapse all component properties context menu support

tkgdhughes:
### Description

Added an option to the context menu which will collapse all the properties of a component. At TKG we have numerous large scripts with deeply nested properties which can be overwhelming to work with. This feature was developed to allow designers using the editor to collapse everything and reduce visual clutter so they could focus on the important attributes of the component.

collapseproperties


ranilianaws:
Hey @tkgdhughes ,

PR #159, opened 2018-05-17
89) Add EntityId_FromString utility function for EntityId.

tkgdhughes:
### Description 

As title, a trivial addition to add a utility function to EntityId.

### DIFF
--- a/dev/Code/Framework/AzCore/AzCore/Component/Entity.cpp
+++ b/dev/Code/Framework/AzCore/AzCore/Component/Entity.cpp
@@ -975,6 +975,25 @@ namespace AZ
     }
 
     //=========================================================================
+    // EntityId_FromString
+    //=========================================================================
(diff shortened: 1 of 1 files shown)


ranilianaws:
Hey @tkgdhughes ,

ranilianaws:
Hey @tkgdhughes ,

Can you tell us more about the use case? Why not use the exposed entity ID type instead?

tkgdhughes:
Hey @ranilianaws, 

I've just been digging back through the usage and I _think_ that this code was written prior to EntityId having a GridMate marshaler. I imagine there was a string marshaler available, so the EntityId which needed sending was forced ToString then back FromString at the other end. I'm wondering if there is any useful use case for FromString anymore now that there is a valid EntityId marshaler. I suppose just from an API symmetry point of view there is a ToString, so a FromString may be useful somewhere.

ranilianaws:
Hey @tkgdhughes ,

Thanks for the contribution but we will not take this change as we don't see a reason to support this functionality.

PR #166, opened 2018-05-23
95) Fix for dedicated server not launching/waiting for the asset processor

tkgdhughes:
### Description 

Fix for the dedicated server not launching or handshaking with the Asset Processor, resulting in it booting up before critical files are copied. If the asset processor is not meant to run it could be disabling via the config file.

- Marked `.ini`, `.cfg`, `.xml`, and `.json` files as critical, so that they are copied before the AP handshake completes. All of these types are used by critical system configuration and therefore needed immediately on boot.
- When launching the dedicated server, it will handshake with (or launch) the non-dedicated version of the asset processor. There isn't a use case for running the version in the dedicated server bin folder, and it fails to start.

### DIFF
--- a/dev/AssetProcessorPlatformConfig.ini
+++ b/dev/AssetProcessorPlatformConfig.ini
@@ -341,6 +341,7 @@ params=copy
 [RC cfg]
 glob=*.cfg
 params=copy
+critical=true
 productAssetType={9952F1E9-BF9F-4B0D-8627-D973A3D82230}
 version=2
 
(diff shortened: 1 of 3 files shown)


ranilianaws:
Hey @tkgdhughes ,

ranilianaws:
Hey @tkgdhughes ,

Thank you for the contribution but we will not be taking this change as is. We will be investigating ways to improve the workflow in the future. Thanks!

PR #167, opened 2018-05-23
96) Add OnUnloadStart trigger

tkgdhughes:
### Description

Add an `OnUnloadStart` function to `ILevelSystemListener` to allow systems to respond to the unloading of a level as appropriate.

### DIFF
--- a/dev/Code/CryEngine/CryAction/ILevelSystem.h
+++ b/dev/Code/CryEngine/CryAction/ILevelSystem.h
@@ -194,6 +194,8 @@ struct ILevelSystemListener
     virtual void OnLoadingError(ILevelInfo* pLevel, const char* error) {}
     //! Called whenever the loading status of a level changes. progressAmount goes from 0->100.
     virtual void OnLoadingProgress(ILevelInfo* pLevel, int progressAmount) {}
+    //! Called before a level is unloaded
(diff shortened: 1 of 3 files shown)


ranilianaws:
Hey @tkgdhughes ,

tkgdhughes:
Morning @ranilianaws,
Unfortunately not, we have both diagnostic and utility code triggering from both OnUnloadStart and OnUnloadComplete.

ranilianaws:
Hey @tkgdhughes ,

We can not accept this as is as we are moving away from CryLegacy event listeners and instead would like to leverage EBuses. I will close this pull request for now but feel free to resubmit the pull request using EBuses. Thank you!

tkgdhughes:
Morning @ranilianaws,

Sorry for the slow response, I was away a couple of days last week! 

Until you guys actually replace the CryLegacy code and re-implement all of it in EBus, does it not make sense to take this change? We use a lot of ILevelSystemListeners in our game and I think it would be useful for others as a stop gap until this fairly integral system is replaced. If there was already an EBus replacement for this I would be happy to add this as an addition but I can't recall one at this moment...?

ranilianaws:
We considered it but we decided not to support any more legacy event listeners. However, we will review an EBus alternative so that users who need this functionality can only import the Ebus-related headers and not CryCommon and so forth.

PR #169, opened 2018-05-24
98) Fix deserialisation of data on linux.

tkgdhughes:
### Description 

Force enum EVoxelGIMode to be a signed int: `Type {72039442-EB38-4D42-A1AD-CB68F7E0EEF6}`, instead of an unsigned int :`Type {43DA906B-7DEF-4CA8-9790-854106D3F983}`. 
This fixes potential serialisation warnings during loading on Linux.

```
Serialize: Trace::Error
/game01/dev/BinTemp/uber_files/AzCore/../../../Code/Framework/AzCore/AzCore/Serialization/SerializeContext.cpp(1867): 'void AZ::SerializeContext::ErrorHandler::ReportError(const char *)'
Serialize: Element 'VoxelGIMode'(0xf5c72d4b) in class 'LightConfiguration' is of type {72039442-EB38-4D42-A1AD-CB68F7E0EEF6} but needs to be type {43DA906B-7DEF-4CA8-9790-854106D3F983}.
```

### DIFF
--- a/dev/Code/CryEngine/CryCommon/IEntityRenderState.h
+++ b/dev/Code/CryEngine/CryCommon/IEntityRenderState.h
@@ -356,7 +356,7 @@ struct IRenderNode
     virtual uint8 GetSortPriority() { return 0; }
 
     // Types of voxelization for objects and lights
-    enum EVoxelGIMode : int
+    enum EVoxelGIMode : signed int
     {
         VM_None = 0, // No voxelization
         VM_Static, // Incremental or asynchronous lazy voxelization


ranilianaws:
Hey @tkgdhughes ,

ranilianaws:
Hmm which component is serializing this enum? Are you referring to LightComponent? 

Also not really clear on the problem space. Is it that on windows it is serialized to int while on linux it is being serialized to unsigned int?  

From the standard:
"If the underlying type is not fixed, the value is convertible to the first type from the following list able to hold their entire value range: int, unsigned int, long, unsigned long, long long, or unsigned long long." 

Which makes me think that it should default to int for both cases as the container is small.

ranilianaws:
@tkgdhughes , any updates on this? Thanks!

tkgdhughes:
Hey @ranilianaws, 

Yes, it was the LightComponent and yes on windows it is serialized to int while on Linux it is serialized to unsigned int.
However, it seems that this isn't actually necessary anymore and I somehow missed that the enum had already been given a fixed type. When the change was originally made the enum did not have a fixed type whereas it does now. Apologies!

PR #171, opened 2018-05-29
101) Add a tick callback to allow injection of code between entity deserialisation

tkgdhughes:
### Description

Similarly to the callbacks `InplaceLoadRootInfoCB`, `ClassReadyCB`, and `CompletionCB`, we have added a `TickCB` to allow injection of code between entity deserialisation. 
This is currently only used from `CLevelSystem::LoadLevelInternal`. In `LoadLevelInternal` we have added a lambda which will pump the loading screen during the synchronous level loading. This was needed because in DRG were seeing stalls due to our large number of entities necessary in the biggest level we used.


ranilianaws:
Hey @tkgdhughes ,

swebrett:
Hey @tkgdhughes .

We've decided not to accept this specific implementation, but you did reveal a shortcoming of our entity deserialization system that needs to be addressed.  We'll review the architecture, and have a solution to this issue in a future version of Lumberyard. Thank you for your contribution!

PR #172, opened 2018-05-29
102) Add CVar to skipSleepIfNeeded

tkgdhughes:
### Description 

Added cvar `sys_ds_skipSleep`. This is a trivial change to cause the dedicated server to not schedule a sleep if it is set to `1`. Turning this on was useful for profiling dedicated server spikes.


ranilianaws:
Hey @tkgdhughes ,

ranilianaws:
Hey @tkgdhughes ,

We reviewed the code and have decided not to take it as we don't recommend disabling the sleep behavior. Thanks for the contribution!

PR #174, opened 2018-05-30
105) Add activate property to camera.

tkgdhughes:
### Description

Added an "Auto-activate" property to the CameraComponent. This allowed us to have multiple camera components enabled and being manipulated in the level without them automatically displaying to the viewport.


ranilianaws:
Hey @tkgdhughes ,

ranilianaws:
Hey @tkgdhughes ,

Thanks for the contribution. We have accepted the code change and it will make it into a future release of LY.

PR #175, opened 2018-05-30
104) Add parameter to optionally load a canvas if it was not found.

tkgdhughes:
### Description 

Added `UiCanvasLuaProxy::BusDisconnect`. This is a deprecated system but as we were still using it, it was useful to have both `BusConnect` and `BusDisconnect` available to Lua.

Added a default parameter `loadIfNotFound = false` to `FindCanvasByPathname`. Our use case for this was to lazily load reticle canvases. This change was initially made to `UiCanvasLuaProxy` but has also been added to the newer `UiCanvasManager`.


gFunc:
Good call out regarding the deprecation, but seems like this is useful 👍 
I'll get this reviewed and let you know how it turns out!

tkgdhughes:
Morning @gFunc,

Thanks for the response. I won't reply to say thanks on all of the pull requests as it gets quite noisy when I start doing that, but I am reading all the comments and will be responding to stuff as necessary!

PR #177, opened 2018-05-30
107) Fix for double checking of error condition

tkgdhughes:
### Description

Simple fix, as title: code in  #`if(!statement)` preceding `AZ_Error` had already checked the validity of the error condition, then it was checking it again unnecessarily,


AMZN-alexpete:
Hey @tkgdhughes,

PR #180, opened 2018-05-31
110) Fix for rain and snow gems.

tkgdhughes:
### Description 

Bug fixes for the snow and rain gems. These are both old systems but are currently the most straightforward way to get rain or snow into a level.

- Both gems were dependent on having an Actor present, but actors are deprecated and we don't use one, so the effects were never triggering.
- The renderer was using the incorrect function to retrieve the camera position for calculating occlusion.


ranilianaws:
Hey @tkgdhughes ,

tkgdhughes:
Hey @ranilianaws,
As far as I am aware there was no real reason other than personal preference for the artist who was working with these entities. I can easily remove that code if you want?

ranilianaws:
It's alright @tkgdhughes, I was just curious. We accepted the code change and it will be in a future release of LY. Thanks for the contribution!

PR #182, opened 2018-06-04
112) Fixes for footstep component.

tkgdhughes:
### Description

- Filter out surface id -1. This id means that the entity was not in contact with the ground.
- In non-release builds, make sure that we don't spam the log output by only warning once per surface.


ranilianaws:
Hey @tkgdhughes ,

AMZN-alexpete:
Hi @tkgdhughes, just wanted to follow up and let you know this fix has been accepted and will be in an upcoming release of Lumberyard.  Thank you!

PR #183, opened 2018-06-04
113) Adding ICameraZoomBehavior interface

tkgdhughes:
### Description

- Augmented the camera rig framework to add a final step which modifies the camera zoom.
- Changed the ICameraTargetAcquirer interface to return the target entity id.

This addition is intended as an example which users can study as a starting point to build their own custom zoom behaviors. For example, DRG used shoulder zooms, a weapon scopes, etc. Multiple behaviors can be combined as would be required for a dynamic third-person camera rig. 

Provided, is a basic DefaultZoomBehavior which can be set to start on activating, and will change the FoV to zoom to the "Zoom Target" over "Time to Zoom" duration. Examples of usage might be for zooming out on activating a player camera, zooming in when changing to a security camera, and so on.

### DIFF
--- a/dev/Gems/CameraFramework/Code/CameraFramework.waf_files
+++ b/dev/Gems/CameraFramework/Code/CameraFramework.waf_files
@@ -4,7 +4,8 @@
             "Include/CameraFramework/ICameraSubComponent.h",
             "Include/CameraFramework/ICameraLookAtBehavior.h",
             "Include/CameraFramework/ICameraTargetAcquirer.h",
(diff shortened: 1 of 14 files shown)


AMZN-alexpete:
Hey @tkgdhughes,

Thank you for the feature. We will evaluate the code and get back to you soon.

AMZN-nggieber:
@tkgdhughes, Thank you! Your feature was accepted and will be in a future LY release.

PR #184, opened 2018-06-05
114) Fix for potential bad raycast

tkgdhughes:
### Description

Fix for a crash seen once in `CPhysicalWorld::RayWorldIntersection`. 

Don't do the part of the raycast which relies on the entity grid if the entity grid is in an invalid state. We caught `m_gthunks` as `nullptr` which caused a crash as it is needed for all calls to `egc.check_cell` including the call to `DrawRayOnGrid`. 

Unfortunately, we did not get the root cause of this, but this should stop any invalid access issues in the future.


ranilianaws:
Hey @tkgdhughes ,

ranilianaws:
Hey @tkgdhughes ,

We have accepted the code and it will be included in a future release of LY. Thanks for the contribution.

PR #185, opened 2018-06-05
115) Load a colour previously serialised to a string

tkgdhughes:
### Description

Add functionality for loading a colour previously serialised to a string.


gFunc:
Thanks for the color helper; I'll make sure this goes up for review... stay tuned!

PR #186, opened 2018-06-05
116) Reflect UiTextComponent::GetTextSize

tkgdhughes:
### Description

Trivial change to reflect `UiTextComponent::GetTextSize`. 

This was added so that we could resize UI elements depending on content (dynamic text in this case). As of LY1.11, this is not as necessary due to the addition of the UiLayoutFitter component, however, it may still have some use.


ranilianaws:
Hey @tkgdhughes ,

ranilianaws:
Hey @tkgdhughes ,

Thanks for the contribution. We have accepted the code and it will be included in a future release of LY.

PR #192, opened 2018-07-25
Fix for V622. Consider inspecting the 'switch' statement.

tkgdhughes:
**Fix for V622. Consider inspecting the 'switch' statement. It's possible that the first 'case' operator is missing.** 

The analyzer has detected a potential error: the first operator in the 'switch' operator's block is not the 'case' operator. It causes the code fragment never to get control.
In this case, it will result in an AZ_Error being missed which would catch incorrect code flow further up the call stack.


gFunc:
This fix has been accepting into Lumberyard with some modifications.  Turns out we don't actually need that method, so it's been removed entirely.  Thanks!

PR #193, opened 2018-07-25
V779. Unreachable code detected. It is possible that an error is present.

tkgdhughes:
This pull request address error code V779 and addresses two separate issues.

**Issue Key: LY-84688 
Issue ID: 203198**
Programmer intended context.m_createdData.push_back(); to be called but due to early returns from the function, it is not ever executed.

**Issue Key: LY-84689 
Issue ID: 203199** 
emit SignalFocused(this); was added in LY 1.9 and it appears as if the programmer wanted this signal to be executed on exiting the Init(...) function. However, due to the function returning before this point, the code will never get executed. By introducing a local boolean parameter it is possible to maintain the same flow of execution whilst also emitting the signal as per the intended design.


tkgdhughes:
Closing pull request, it was meant to be against 1.14, not master

PR #195, opened 2018-07-25
V611. The memory allocation and deallocation methods are incompatible.

tkgdhughes:
**Issue Key: LY-84687 
Issue ID: 203197** 

The memory was allocated using 'new T[]' operator but was released using the 'delete' operator. Using delete on a pointer returned by new [] or delete [] on a pointer returned by new results in undefined behaviour.


SRugina:
I also also caught that bug, which is one of the commits in my pull request

gFunc:
Fixed accepted into Lumberyard!  Thank you!

PR #196, opened 2018-07-25
V549. The 'first' argument of 'Foo' function is equal to the 'second' argument.

tkgdhughes:
**Issue Key: LY-84686 
Issue ID: 203196** 

The 'first' argument of 'Foo' function is equal to the 'second' argument.

Due to what is probably a typo, the implementation of using `memcpy` to compare the size of these two vectors will always return true as the two first arguments are the same.


SRugina:
I also caught that bug, which is one of the commits in my "Patches for 1.14.0.1" pull request

gFunc:
This fix has been accepting into Lumberyard proper.  Thanks!

PR #197, opened 2018-07-25
V596. The object was created but it is not being used. The 'throw' keyword could be missing.

tkgdhughes:
Two separate issue fixed in one commit:

**Issue Key: LY-84685** 
**Issue ID: 203195**

**Issue Key: Y-84684** 
**Issue ID: 203194**

The analyzer has detected a strange use of the std::exception class or derived class. The analyzer generates this warning when an object of the std::exception / CException type is created but not being used. In this case, a std::runtime_error(...) is being created but not thrown.


SRugina:
I also caught that bug, which is one of the commits in my "Patches for 1.14.0.1" pull request

PR #198, opened 2018-07-25
Patches for 1.14.0.1

SRugina:
Hi!

I decided to clean up some of the errors and typos in the code. Some of these commits are only comments and not actual changes, since it was unclear what the author had intended.

### DIFF
--- a/dev/Code/CryEngine/CryAnimation/ModelMesh.cpp
+++ b/dev/Code/CryEngine/CryAnimation/ModelMesh.cpp
@@ -171,7 +171,7 @@ ClosestTri CModelMesh::GetAttachmentTriangle(const Vec3& RMWPosition, const Join
         Vec3 v1 = *(Vec3*)(pPositions + pIndices[d + 1] * nPositionStride);
         Vec3 v2 = *(Vec3*)(pPositions + pIndices[d + 2] * nPositionStride);
         Vec3 TriMiddle = (v0 + v1 + v2) / 3.0f + m_vRenderMeshOffset;
-        f32 sqdist = (TriMiddle - RMWPosition) | (TriMiddle - RMWPosition);
+        f32 sqdist = (TriMiddle - RMWPosition) | (TriMiddle - RMWPosition); //(TriMiddle - RMWPosition) duplicated, possible typo
         if  (distance > sqdist)
         {
             distance = sqdist, foundPos = d;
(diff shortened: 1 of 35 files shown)


SRugina:
A thing to note is that some of my "temporary fixes" and "possible typos" have been properly fixed by other pull requests

tkgdhughes:
Great stuff @SRugina

SRugina:
Yeah I noticed you've fixed some of the things I wasn't sure about, which is great!

tkgdhughes:
I have a load more coming too so watch this space.

dev/Code/CryEngine/CryAnimation/ModelMesh.cpp
```diff
@@ -171,7 +171,7 @@ ClosestTri CModelMesh::GetAttachmentTriangle(const Vec3& RMWPosition, const Join
         Vec3 v1 = *(Vec3*)(pPositions + pIndices[d + 1] * nPositionStride);
         Vec3 v2 = *(Vec3*)(pPositions + pIndices[d + 2] * nPositionStride);
         Vec3 TriMiddle = (v0 + v1 + v2) / 3.0f + m_vRenderMeshOffset;
-        f32 sqdist = (TriMiddle - RMWPosition) | (TriMiddle - RMWPosition);
+        f32 sqdist = (TriMiddle - RMWPosition) | (TriMiddle - RMWPosition); //(TriMiddle - RMWPosition) duplicated, possible typo
```
tkgdhughes: I have just been looking at this one and wondered if you had addressed it. It had me confused at first, but the code is actually correct. The `|` operator is overloaded to perform the dot product on two Vec3s. If the vectors are parallel then the dot product just results in the ordinary multiplication of their lengths. If the vectors were normalised then this would just result in 1, but as they are not we get the squared distance. It's not clear, but it is correct!

tkgdhughes: Haha yeah, I am very much a fan of optimizing for the reader. The compiler is pretty smart whereas mortal programmers get confused easily! It took me longer than I am willing to admit to realise what that code was doing ;)

SRugina:
seeing as these issues are getting covered more competently by others, this pull request is not required

gFunc:
@SRugina, thanks for the contribution!  Lumberyard has accepted a few of these fixes already :D 
I'll double check to make we didn't miss anything.

PR #199, opened 2018-07-26
V564. The '&' or '|' operator is applied to bool type value. You've probably forgotten to include parentheses or intended to use the '&&' or '||' operator.

tkgdhughes:
4 fixes in this pull request related to code V564, I'll comment on the changes individually.

### DIFF
--- a/dev/Code/CryEngine/RenderDll/XRenderD3D9/D3DHWShader.cpp
+++ b/dev/Code/CryEngine/RenderDll/XRenderD3D9/D3DHWShader.cpp
@@ -1828,8 +1828,8 @@ namespace
                 {
                     if (gcpRendD3D->FX_GetEnabledGmemPath(nullptr))
                     {
-                        uint32 stencilRef = CRenderer::CV_r_VisAreaClipLightsPerPixel ? 0 : (rRP.m_RIs[0][0]->nStencRef | BIT_STENCIL_INSIDE_CLIPVOLUME);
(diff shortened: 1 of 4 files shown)


dev/Code/CryEngine/RenderDll/XRenderD3D9/D3DHWShader.cpp
```diff
@@ -1828,8 +1828,8 @@ namespace
                 {
                     if (gcpRendD3D->FX_GetEnabledGmemPath(nullptr))
                     {
-                        uint32 stencilRef = CRenderer::CV_r_VisAreaClipLightsPerPixel ? 0 : (rRP.m_RIs[0][0]->nStencRef | BIT_STENCIL_INSIDE_CLIPVOLUME);
-                        stencilRef |= (!(rRP.m_pCurObject->m_ObjFlags & FOB_DYNAMIC_OBJECT) | CRenderer::CV_r_deferredDecalsOnDynamicObjects ? BIT_STENCIL_RESERVED : 0);
+                        uint32 stencilRef = CRenderer::CV_r_VisAreaClipLightsPerPixel ? 0 : (rRP.m_RIs[0][0]->nStencRef | BIT_STENCIL_INSIDE_CLIPVOLUME);                        
+                        stencilRef |= (!(rRP.m_pCurObject->m_ObjFlags & FOB_DYNAMIC_OBJECT) || (CRenderer::CV_r_deferredDecalsOnDynamicObjects ? BIT_STENCIL_RESERVED : 0));
```
tkgdhughes: This is another simple case of using the `|` operator instead of the `||` operator.
From examining the code it is fairly clear that the intention is that if the `FOB_DYNAMIC_OBJECT` flag is set, then only set `stencilRef` if `CV_r_deferredDecalsOnDynamicObjects` is non-zero. Previously, the code would never set `stencilRef` if the 'FOB_DYNAMIC_OBJECT' flag is set.

PR #202, opened 2018-07-26
V561. It's probably better to assign value to 'foo' variable than to declare it anew.

tkgdhughes:
There are a number of commits here all relating to variables being declared multiple times and potentially causing unwanted behaviour. See the commits for discussions of each change.


gFunc:
The Editor team would like to test the changes in QGradientSelectorWidget.cpp; we agree with the fix but want to make sure that something else happened to work because QGradientSelectorWidget was broken.  I'm going to investigate this a bit further and let you know how it goes :)

PR #205, opened 2018-07-26
V546. Member of a class is initialized with itself: 'Foo(Foo)'.

tkgdhughes:
This looks like a simple intellisense autofill error, EConfigSpecBrief was meant but the programmer tabbed out to get eConfigMax.


gFunc:
This fix has been accepted into Lumberyard.  Thanks!

PR #215, opened 2018-07-27
CZBufferCuller: There are identical sub-expressions 'MaxX < 0' to the left and to the right of the '||' operator.

tkgdhughes:
**Issue key: LY-84628 
Issue id: 203108**

**Trivial error / code cleanup:**
Fix typo `MaxX` which should be `MaxY`. This file doesn't even seem to be used anymore and could possibly just be removed entirely.

### DIFF
--- a/dev/Code/CryEngine/Cry3DEngine/CZBufferCuller.h
+++ b/dev/Code/CryEngine/Cry3DEngine/CZBufferCuller.h
@@ -125,7 +125,7 @@ protected:
         }
         if (ROTATE == 2)
         {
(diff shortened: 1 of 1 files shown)


gFunc:
This one is strange.  MaxX and MaxY can never be less than 0, so there might be more problems here than meet the eye, ha!

AMZN-alexpete:
Thanks for submitting this fix, this code is still used by the coverage buffer, but the e_CoverageBufferRotationSafeCheck defaults to zero so this code is not normally used.  Gene is also correct that this code should never be used, but fewer compiler warnings means less noise so we're integrating the change into a future version of Lumberyard.

PR #220, opened 2018-07-27
ShaderComponent: V501. There are identical sub-expressions to the left and to the right of the 'foo' operator.

tkgdhughes:
**Issue key: LY-84623 
Issue id: 203100**

**Issue key: LY-84622 
Issue id: 203099**

**Code cleanup:**
Removing identical sub-expressions in != operator, it is a fairly common and (mostly) benign pattern in equality and in-equality operators to accidentally duplicate checks


gFunc:
Your fix has been adopted into Lumberyard, thanks!

PR #222, opened 2018-07-27
Texture: There are identical sub-expressions '(m_eTFSrc == eTF_BC6UH)' to the left and to the right of the '||' operator

tkgdhughes:
**Issue key: LY-84643 
Issue id: 203133** 

**Bug fix:**
This was a tricky one as I am not particularly familiar with the different possible texture formats, so I fell back to searching the codebase for similar functionality and comparing the flagged code with other instances doing the same thing. Fortunately, `Image_DXTC.cpp` is doing very similar checks for base range normalisation and we can see on line 601, 629, and 645 that the intended texture format enum is actually `eTF_BC6SH`. It is easy to see how this typo could slip past the implementing programmer and the reviewer. The result of this error will be that textures with the `eTF_BC6SH` format will not be normalized correctly.


gFunc:
This is now up for review... great detective work!  Thanks!

PR #228, opened 2018-08-01
Budgeting System: Several cases exist where the else and then statements are equivalent

tkgdhughes:
**Issue key: LY-84619 
Issue id: 203092** 

Cleaning up a benign error in an old CrySystem. 

I'm not sure how this error ended up here, but we could speculate that maybe in the block for `if(scale <= 1.0f)`, `DrawText` used to be called with one color, then in the else block it was called with a different one. The programmer may have refactored the code to use the `GetColor` function and not tidied up fully.


AMZN-nggieber:

PR #230, opened 2018-08-01
LivingEntity: several cases exist where the else and then statements are equivalent

tkgdhughes:
**Issue key: LY-84619 
Issue id: 203092**

**Code cleanup**
V523. The 'then' statement is equivalent to the 'else' statement.
Again we can see equivalent if/else blocks rendering the conditional unneeded. It is likely that in the eST_Network case the code used to do something different (as seen in the commented out code), but no longer does. If it is the intention of the programmer to refer to this code then source control will still contain the details for them.


AMZN-nggieber:

PR #234, opened 2018-08-01
TrackViewNodes: several cases exist where the else and then statements are equivalent

tkgdhughes:
**Issue key: LY-84619 
Issue id: 203092 **

**Bug fix**
V523. The 'then' statement is equivalent to the 'else' statement.

Copy and paste error. The programmer has clearly copied and pasted the if block into the else blocks throughout the file, and then done a pass to update the else block to do the correct non-legacy functionality. They obviously missed this one.


AMZN-nggieber:
@tkgdhughes Thank you for the submission, code has since been refactored so issue is no longer present.

PR #237, opened 2018-08-01
AttributePool: Fix for memory leak and for "expression always true" error.

tkgdhughes:
**Bug fix / Code cleanup**

Fixes for:
V547 Expression 'poolType == POOLTYPE_DYNAMIC' is always true. attributepool.cpp 179
V773 The function was exited without releasing the 'pool' pointer. A memory leak is possible.

More notes with the individual commits.


AMZN-nggieber:
@tkgdhughes thank you for these fixes, the attribute pool has since been deprecated.

PR #239, opened 2018-08-01
AssetDatabase: V1004. The pointer was used unsafely after it was verified against nullptr

tkgdhughes:
**Code cleanup**

V1004. The pointer was used unsafely after it was verified against nullptr
The analyzer has detected a possible null-pointer-dereferencing error. The pointer is checked for null before the first use but is then used for the second time without such a check.

Fixing a number of potential null pointer dereferencing errors. This file was flagged as a high-risk file due to the presence of a number of high priority issues alongside the fact that it is not a legacy Cry file. A number of the issues actually turned out to be benign and have been silenced in the analyzer, this error, however, is worth fixing.


AMZN-nggieber:
@tkgdhughes Thanks for the improvements, code has been refactored since to remove the issues

PR #244, opened 2018-08-03
ImageObject: V773 The function was exited without releasing the 'pAttached' pointer. A memory leak is possible.

tkgdhughes:
This is a simple fix for a possible memory leak in ImageObject.


AMZN-nggieber:
@tkgdhughes thank you for the suggestion, ResourceCompilerImage has been deprecated and removed now

PR #246, opened 2018-08-03
V560. A part of conditional expression is always true/false.

tkgdhughes:
Code cleanup, remove an unnecessary check of datum after it has already been verified.


AMZN-nggieber:
@tkgdhughes thank you for the suggestion. This code has since been changed so this is no longer relevant, closing

PR #253, opened 2018-08-06
boolean3d: V773. The function was exited without releasing the pointer/handle. A memory/resource leak is possible.

tkgdhughes:
Due to an early return midway through this function, it is possible to leak the memory allocated by `removedTri` and `newTri`.


AMZN-nggieber:

PR #255, opened 2018-08-06
CharacterManager: V773. The function was exited without releasing the pointer/handle. A memory/resource leak is possible.

tkgdhughes:
If `numChildren == 0`, the memory allocated for `def` will leak. Fix by deleting `def` correctly.


AMZN-nggieber:

PR #259, opened 2018-08-06
HeightFieldGeom: V773. The function was exited without releasing the pointer/handle.

tkgdhughes:
A memory/resource leak is possible. Clean up allocated memory and move destructor into source file as it is no longer trivial.


AMZN-nggieber:

PR #260, opened 2018-08-07
BreakablePlane: V773. The function was exited without releasing the pointer/handle. A memory/resource leak is possible.

tkgdhughes:
Early return from function causes a memory leak as pPlane is not deleted.


AMZN-nggieber:

PR #266, opened 2018-08-07
LoaderTCB: V773. The function was exited without releasing the pointer/handle. A memory/resource leak is possible.

tkgdhughes:
Fix for a number of early returns leading to memory leaks.


AMZN-nggieber:

PR #270, opened 2018-08-07
Geometry: V778. Two similar code fragments were found. Perhaps, this is a typo and 'X' variable should be used inst

tkgdhughes:
**Bug fix**

From inspection of the code fragment we can see that the intention is to swap `iPrim[0]` and `iPrim[1]` but the programmer has clearly copy pasted the code fragment here... 

```
        i = gtest.contacts->id[0];
        gtest.contacts->id[0] = gtest.contacts->id[1];
        gtest.contacts->id[1] = i;
```

... and not swapped out the variables correctly. Looking elsewhere in this file can also help to confirm that this is what was intended.


AMZN-nggieber:

PR #278, opened 2018-08-09
AIMemStats: V568 It's odd that 'sizeof()' operator evaluates the size of a pointer to a class, but not the size of the 'vol' class object

tkgdhughes:
**Bug Fix**

The argument of the sizeof() operator is a pointer to a class. In most cases this shows that the programmer forgot to dereference the pointer. In this case it is clear that vol should be dereferenced so that we are returning the size of the object, not the pointer, for the MemStats.

This is probably a deprecated class, but if someone copies this code the bug will propagate.


AMZN-nggieber:
@tkgdhughes thank you for this suggestion, the code has since changed but we reviewed for any similar errors as this potentially still in the code.

PR #281, opened 2018-08-13
Physinterface: V519. The 'x' variable is assigned values twice successively. Perhaps this is a mistake.

tkgdhughes:
This file had a lot of errors flagged in it but many turned out to be false positives or to be strange but correct and probably not worth the risk of fixing. These two minor changes are what was left; simple cases of double assignment without usage in between each assignment.


AMZN-nggieber:

PR #297, opened 2018-08-16
Fix black outlines for r_AntialiasingMode 0/1

alex-leleka:
Objects gain black outlines if Anti-aliasing is set to 0 or 1. 
This happens due to TAA_SHARPENING_FACTOR > 1.0f in lerp( cFiltered, OUT.Color.rgb * OUT.Color.rgb, TAA_SHARPENING_FACTOR)
shader PostAAComposites_PS

### DIFF
--- a/dev/Code/CryEngine/RenderDll/XRenderD3D9/GraphicsPipeline/PostAA.cpp
+++ b/dev/Code/CryEngine/RenderDll/XRenderD3D9/GraphicsPipeline/PostAA.cpp
@@ -588,7 +588,7 @@ void PostAAPass::RenderComposites(CTexture* sourceTexture)
     }
     else
     {
-        const Vec4 temporalParams(0, 0, 0, max(1.0f + CRenderer::CV_r_AntialiasingTAASharpening, 1.0f));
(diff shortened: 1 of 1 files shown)


gFunc:
Thanks for the pull request!  I'll follow up with the graphics team and keep you posted :)

AMZN-alexpete:
Hi @alex-leleka, in Lumberyard we had a different fix internally for this issue which was to change the cvar name to be CV_r_Antialiasing**Non**TAASharpening which more accurately describes what the sharpening is for.  This cvar has a default value of 0 (no sharpening) and leaves the original max as it was in the CryEngine implementation.  CV_r_AntialiasingTAASharpening was really only ever supposed to be used with mode 2 (SMAA-TAA) - see http://www.crytek.com/download/Sousa_Graphics_Gems_CryENGINE3.pdf slide 19 - and was not supposed to be active in modes 0 or 1, though it can be used to sharpen the image if you want that effect in a single shader instead of activating r_sharpening which is a separate post-process effect.

Mode 3 (TAA) does sharpening in a different step when looking up the history buffer.

Do you plan on using this sharpening in modes 0 or 1?

alex-leleka:
Hello @AMZN-alexpete. Thank you for detailed response. 
We are ok with default Lumberyard value for this parameter. We don't have a need for sharpening in modes 0 or 1.

PR #304, opened 2018-08-26
 Translate the Sandbox user interface into Simplified Chinese

GameInstitute:
Identify the language type via QLocale::system().name().toLower().
For example, I'd used zh_cn for Simplified Chinese.


swebrett:
Hey, @GameInstitute .

Thanks for the pull request!  We'll take a look at the code (and the translations) and get back to you soon.

PR #307, opened 2018-09-10
Linux mismatch free and delete calls

rustamserg:
Valgrind detects that due to clang optimization there is mismatch between free and delete calls. The clang optimization should be also disabled for Linux as well.


AMZN-puvvadar:
Hi @rustamserg ,

I've started evaluation of this. Do you have any testing steps we could use to confirm this one given it's in memory management?

rustamserg:
We found this problem from Valgrind log. Right now I don't have logs however I will try revert the change and reproduce it once again.

rustamserg:
[output.log](https://github.com/aws/lumberyard/files/2492270/output.log)

Please find attached log file captured from Valgrind session with reproduced memory problem.

Please find below line in the log to see details:
**Mismatched free() / delete / delete []**

AMZN-puvvadar:
Hi @rustamserg ,

Thanks for the data, it was very helpful. While our implementation will look slightly different, we've accepted this change. Thanks for the submission!

PR #308, opened 2018-09-10
Crash on cry memory init

rustamserg:
Cry memory calls can crash when two threads in one module try to allocate the memory at the same time. The initialization flag is set up too early.


AMZN-puvvadar:
Hi @rustamserg ,

We'll begin evaluation of this as soon as we complete evaluation on the Linux free/delete mismatch submission.

AMZN-puvvadar:
Hi @rustamserg ,

We won't be able to take this change as the affected code has changed significantly. These changes should also address the issue, thanks for the submit!

PR #309, opened 2018-09-10
Missed on thread exit bus call

rustamserg:
RAD telemetry tool incorrectly detects when cry thread is exit due to missed bus call.


AMZN-puvvadar:
Hi @rustamserg ,

Apologies for the delay, this request has been accepted. Thanks again for the submission!

PR #310, opened 2018-09-11
Fix linux cases

rustamserg:
Fix linux case sensitive dedicated server configuration build


AMZN-puvvadar:
Hi @rustamserg ,

Thanks for the submission, we've begun review on it. I imagine this will be fairly straightforward.

AMZN-puvvadar:
Hi @rustamserg ,

PR #311, opened 2018-09-11
Do not create qtlibs folder if no qt is available

rustamserg:
On linux dedicated server we create unnecessary empty qtlibs folder


AMZN-puvvadar:
Hi @rustamserg,

Apologies for the delays. I've started review on this and will be going through the recent bulk of reviews submitted. Thanks again for all of these.

AMZN-puvvadar:
Hi @rustamserg ,

PR #312, opened 2018-09-11
Support vcvars args configuration for VS2017

rustamserg:
Support using --win-vs2017-vcvarsall-args default setting. Now it can be configured as follows:
 "default_value": "-vcvars_ver=14.13" to stick with specific vs version


AMZN-puvvadar:
Hi @rustamserg ,

Thanks for the change, we've started evaluation of it.

AMZN-puvvadar:
Hi @rustamserg ,

We've accepted this change. Thanks for the submission!

PR #313, opened 2018-09-11
Add distcc support for Linux

rustamserg:
Changes in WAF system required to start using distcc for Linux dedicated server compilation


AMZN-puvvadar:
Hi @rustamserg ,

Are there any other changes made for distcc/are these just the initial changes?

rustamserg:
There are no other changes in original Lumberyard engine to enable distcc compilation for Linux dedicated server. We also disabled precompiled headers to start using distcc but the code is compiled well in this mode except of Kythera which is not part of the engine

AMZN-puvvadar:
Hi @rustamserg ,

We've found these changes to WAF to have value by themselves and have accepted them on that rationale. Thanks for the submission!

PR #314, opened 2018-09-11
Waf ref before assigment

rustamserg:
Fix using reference before assignment
Fix match call on object not string


AMZN-puvvadar:
Hi @rustamserg ,

I've started review on this, thanks!

AMZN-puvvadar:
Hi @rustamserg ,

PR #315, opened 2018-09-11
Support map with string key serialization

rustamserg:
Fix serialization of map container with string type as key

### DIFF
--- a/dev/Code/CryEngine/CryCommon/Serialization/STLImpl.h
+++ b/dev/Code/CryEngine/CryCommon/Serialization/STLImpl.h
@@ -242,7 +242,7 @@ namespace   std
     }
 
     template<class V>
-    bool Serialize(Serialization::IArchive& ar, std::pair<string, V>& pair, const char* name, const char* label)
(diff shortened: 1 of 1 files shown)


AMZN-puvvadar:
Hi @rustamserg ,

Can you describe the original issue here in some more detail? What was the serialization error that was seen?

dkondrashkin:
Hi, @AMZN-puvvadar,

We've encountered an issue with map container deserialization - it did not work for maps using ```string``` type as a key. In this case compiler has always chosen first ```Serialize(...)``` function overload (accepting ```std::pair<K, V>&``` argument) which, in turn, resulted in an empty map instance. Then we discovered that  second overload treats string keys in a different, yet working way. So we've came up with this fix.

AMZN-puvvadar:
Thanks @dkondrashkin ,

We've started review of this submission.

AMZN-puvvadar:
@dkondrashkin @rustamserg ,

We've accepted this change, thanks again for the submission!

PR #317, opened 2018-09-11
Remote console thread safe

rustamserg:
Make remote console thread safe
Fix memory leak on client closing


AMZN-puvvadar:
Hi @rustamserg ,

What's the rationale behind moving the FillAutoComplete function from the client to the server here?

rustamserg:
The problem we are trying to fix is caused by crash in FillAutoCompleteList when many clients are created simultaneously in different threads.

It is reproduced well in out automation testing that heavily uses remote console to orchestrate the test scenario.

Since all the clients share the single instance of the server we decided to move auto complete functionality into the server object and protect it with synchronization primitive.

If we decide to leave it on client then we need something global for all clients to make FillAutocompleteList function thread-safe.

AMZN-puvvadar:
Great, thanks! This will help us evaluate the change faster. We've started the review process on it.

AMZN-puvvadar:
Hi @rustamserg ,

PR #318, opened 2018-09-11
OnAssetReady support changing of queue slice instantiation

rustamserg:
In some cases OnAssertReady starts iterating on container during its modification. Thus we have to use find_if instead of iteration

### DIFF
--- a/dev/Code/Framework/AzFramework/AzFramework/Entity/EntityContext.cpp
+++ b/dev/Code/Framework/AzFramework/AzFramework/Entity/EntityContext.cpp
@@ -731,60 +731,65 @@ namespace AzFramework
             [this, readyAsset]() // we intentionally capture readyAsset by value here, so that its refcount doesn't hit 0 by the time this call happens.
             {
                 const AZ::Data::AssetId readyAssetId = readyAsset.GetId();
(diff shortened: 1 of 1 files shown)


AMZN-puvvadar:
Hi @rustamserg ,

Is there a purpose to the while(true) part of this change? It seems find_if by itself should handle all prior functionality captured by the iterating loop.

AMZN-puvvadar:
Hi @rustamserg ,

We've started evaluation of this. Are there any repro steps you could provide?

rustamserg:
This problem is caused by our content and the way we use asset system. Extract clean repo is really hard however I will try to explain our motivation to share the change:

- OnAssertReady contains iteration over internal container
- During the iteration external source code is called
- The external call can change the internal container on the same callstack during iteration

We can avoid this problem on caller side but we believe that caller should not know about entity context internal logic so we decided to rewrite iteration in the way that stable to container modifications.

AMZN-puvvadar:
Hi @rustamserg ,

Thanks for the details, this is very helpful. We've started review for this.

AMZN-puvvadar:
Hi @rustamserg ,

PR #322, opened 2018-09-17
Enable Lensflares in Voxel GI mode

alex-leleka:
GI mode for Lens Flares is missed so it doesn't work in Voxel GI mode.


swebrett:
Hey, @alex-leleka !

Thanks for the pull request.  We'll take a look at the code, and get back to you soon!

PR #323, opened 2018-09-28
Add Static MeshComponent to SVOGI

alex-leleka:
Add Static MeshComponent render nodes to svo ti so it can voxelize them.

### DIFF
--- a/dev/Gems/SVOGI/Code/Source/SvoBrick.cpp
+++ b/dev/Gems/SVOGI/Code/Source/SvoBrick.cpp
@@ -770,7 +770,8 @@ namespace SVOGI
             for (AZ::s32 nObjType = 0; nObjType < eERType_TypesNum; nObjType++)
             {
                 if ((nObjType == eERType_Brush) ||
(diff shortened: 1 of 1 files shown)


swebrett:
Hey, @alex-leleka !

Thanks for the pull request.  We'll take a look at it, and post an update here once we've done so!

swebrett:
Hi again, @alex-leleka !

We've had a chance to look at the code you submitted, and I was curious if you could provide us with some more context.  Do you have a simple level that shows a change in SVOGI effects before and after this change?  Static entities with Mesh Components should already contribute to SVOGI.  If you don't see those results, though, I'd like to take a look at your SVOGI settings.  What Total Illumination Integration Mode are you using? I see the attached result when using Static Mesh Components with Total Illumination Integration Mode set to 1.

gi on

alex-leleka:
hello @swebrett you are right, mesh components already contributes to GI on empty level it much easier to see. It seems it process levels with complex geometry slower than in 1.14, but that's another talk. 
Thank you. I close pull request.

PR #324, opened 2018-10-12
BeachCity Update

OneManStudios-Pa1:
Would you like to merge BeachCity project into Lumberyard?
I updated BeachCity to work with Lumberyard 1.15, I can give it as a PR or can share it separately as a zip file.


AMZN-alexpete:
Hi @OneManStudios-Pa1, thank you for your pull request, but we do not intend to include BeachCity in current releases of Lumberyard.  It has some great assets, but StarterGame is a better fit for customers who want assets and game logic.

PR #325, opened 2018-11-07
Bugfix: Peer can not activate netsynced entities after disconnecting …

CDufour909:
…from server

Fixes the issue where a peer disconnecting from a server could not automatically activate net-synced entities when loading a new local-only level.


AMZN-puvvadar:
Hi @CDufour909 ,

Thanks for the submission, we've started review of the change.

AMZN-puvvadar:
And we've accepted the change :)

Thanks again, for the submit!

PR #326, opened 2018-11-08
lds cpu usage optimization while no session hosted/level loaded

ximura:
* PhysX , subscribe on Tick Handler after session hosted
* Physx, use default cpu dispatcher to prevent thread starvation for anothers jobs on DS
* EmotionFX,  subscribe on Tick Handler after session hosted
* ScriptSystem, ignore Update while no level loaded
* Create Null realization for Stream Engine
* Disable posteffects update on DS
* Add configuration for threads used by job system and asset manager
** job_thread_count - startup argement to setup amount of threads for job system
**  asset_thread_count -  startup argement to setup amount of threads for asset manager
* bMinimal disable some of systems that isn't server related (ex. VR), but we need to have audio system and gems for correct work

### DIFF
--- a/dev/Code/CryEngine/Cry3DEngine/PostEffectGroup.cpp
+++ b/dev/Code/CryEngine/Cry3DEngine/PostEffectGroup.cpp
@@ -210,7 +210,9 @@ PostEffectGroupManager::PostEffectGroupManager()
     PostEffectGroup* base = new PostEffectGroup(this, "Base", PostEffectGroup::kPriorityBase, true, 0.f);
     base->SetEnable(true);
     m_groups.push_back(std::unique_ptr<PostEffectGroup>(base));
+#ifndef DEDICATED_SERVER
     gEnv->pRenderer->RegisterSyncWithMainListener(this);
+#endif
     if (gEnv->IsEditor())
     {
         //  Only monitor assets in the editor.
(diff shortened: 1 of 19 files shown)


AMZN-puvvadar:
Hi @ximura ,

Thanks for the submission! We've started evaluating the change. It's a rather large change but we'll work to evaluate it as quickly as possible.

AMZN-puvvadar:
Hi @ximura ,

We've broken this into 5 separate changes and will be giving status on each as they become available. 

On the changes to add SessionTickComponent. We've rejected this change as we will be looking at a different way to address this type of situation. We also had some feedback we felt worth sharing.

- Instead of a component, you may want to model encapsulating two handlers similarly to InputChannelEventListener
-You may want to have handling for OnNetworkSessionDeactivated

AMZN-puvvadar:
Hi @ximura ,

Do you have more information on the thread starvation encountered in the physics system? In what situations did you see this? The PhysX job system should run at a fixed rate against available job threads so starvation is concerning. There are concerns that taking physics off of the AZ Jobs system will lead to issues in scenarios with larger simulations so we will likely be unable to take that change but would like to investigate the root cause further.

AMZN-puvvadar:
Hi @ximura ,

On the job thread/asset thread configuration changes. We'll be rejecting this since we don't feel ComponentApplication is the appropriate home for this configuration. Our recommendation is housing those configs at the SystemComponent level (which is something we'll look at).

AMZN-puvvadar:
Hi @ximura,

For the NullRenderer and ScriptSystem  update ignore changes, we'll be accepting both with some minor changes on our end.

PR #327, opened 2018-11-08
Clear cache after Metastream server shutdown

ximura:
Metastream has access to old cache, after shutdown and starting again


AMZN-puvvadar:
Hi @ximura ,

We've started review of this.

AMZN-puvvadar:
Hi @ximura ,

This change has been accepted. Thanks for the submission!

PR #336, opened 2018-12-06
r_stats 6 console command shadow counter fix

SSPkrolik:
**Problem:**
r_stats 6 console command does not show amount of shadow draw calls


AMZN-alexpete:
@SSPkrolik, thank you for submitting this fix!  We have assigned an internal issue and will update this pull request after it has been accepted of if we need more information.  Cheers!

PR #339, opened 2018-12-06
Allow running non-release game client on PC with null renderer

SSPkrolik:
This change allows running client with NULL Renderer on PC which is helpful for us when building command-line utilities within engine's environment.


AMZN-alexpete:
Thanks for submitting this feature @SSPkrolik, we created the issue LY-90249 but forgot to update the actual pull request with the information till now.  Cheers!

AMZN-alexpete:
Hi @SSPkrolik, we added a similar feature back in March but the argument is -NullRenderer and it doesn't set the native width/height.  When I tested it recently it didn't launch a visible window when using the -NullRenderer with the game launcher.  Are you still able to use this feature in a recent (1.25+) version of Lumberyard?

PR #340, opened 2018-12-06
Implemented proper selection box for decals in editor

SSPkrolik:
**Fixes Problem:**
Currently when you have Decals in editor no matter what their size is it is only possible to select those in viewport by directly clicking helper icon and is almost impossible to select it without helpers enabled.

This PR fixes selection box for decals in Editor


AMZN-alexpete:
Thank you for submitting this fix @SSPkrolik,  we've created an internal issue LY-90252 to track it.  Cheers!

PR #341, opened 2018-12-06
Fixed _displ not defaulting to Displacement in RC + new suffixes for terrain albedos

SRugina:
was getting really annoyed having to manually set my _displ textures to Displacement in RC, and having to set my terrain material _diff textures to "Terrain_Albedo_HighPassed" so added the _diffTHP suffix for that.

also added a _diffT suffix for people who don't have Photoshop to highpass the texture, and want to save time not having to manually set the RC to "Terrain_Albedo" (which highpasses the texture for them).


AMZN-alexpete:
Hello @SRugina and thank you for submitting this fix.  We've created an internal issue LY-90278 to track it and will update this PR when it has been submitted.

PR #342, opened 2018-12-07
Fixed CryCommon STLImpl serialization

SSPkrolik:
Fixed CryCommon STLImpl.h Serialization


AMZN-puvvadar:
Rejected as we already have this change from you. :)

See https://github.com/aws/lumberyard/pull/315. Thanks again for the submit!

PR #343, opened 2018-12-07
Fixed WaterVolume component caused crash in Editor and in Game

SSPkrolik:
Fixed crash of watervolume (easily reproduced in editor when entering/exiting game mode several times)


AMZN-puvvadar:
Hi @SSPkrolik ,

Do you have any technical detail on the crash? Were you seeing that the original line would rarely grab an incorrect override/was an incompatible object type being checked?

SSPkrolik:
Hi, @AMZN-puvvadar, I've investigated once more on pure 1.16 official build and found that another change in physics system caused this crash, so everything is okay, closing this one.

PR #344, opened 2018-12-07
Fixed flipped normal when using fillspace designer tool

SSPkrolik:


AMZN-puvvadar:
Hi @SSPkrolik ,

PR #345, opened 2018-12-07
Fixed call of functions declared under #define

SSPkrolik:
ICharacterManager has a bunch of debug functions declared under BLENDSPACE_VISUALIZATION define. 
All these function calls should also be called under this define


AMZN-puvvadar:
Hi @SSPkrolik ,

Sorry for the delay, we've started evaluating this change.

AMZN-puvvadar:
Hi @SSPkrolik ,

PR #347, opened 2018-12-10
Fixed e_coverageBufferDebug rendering artifacts

SSPkrolik:
Right now in engine version 1.16 e_coverageBufferDebug renders cpu filled image which flickers with artifactrs and it actually very hard to use such culling debug.

This fix performs copying cpu filled texture into rendering device texture which produces smooth monochrome (green) texture without flickering and rendering artifacts.


AMZN-alexpete:
Thank you for submitting this fix @SSPkrolik!  We've created an internal issue LY-90345 to track the progress and will update this PR when it has been submitted into a release.

PR #348, opened 2018-12-10
Enabled MeshComponent to be rendered in camera space

SSPkrolik:
Rendering subsystem of The Engine allows meshes to be rendered in camera space. This PR allows to enable this behaviour for mesh component.

This feature is especially important for first person games because it allows to not bother with "weapon-in-a-wall" problem which is solved by physics otherwise.


AMZN-alexpete:
Thanks @SSPkrolik,  I hate it when my weapon gets stuck in a wall :)  We've created an internal issue LY-90347 to track the progress of getting this into a release.

PR #349, opened 2018-12-10
Extended ConsoleBus with methods for getting cvar values

SSPkrolik:
Currently it is possible to set configuration variable from code via ExecuteString method, but it is not possible to retrieve console variable values (without bringing whole CrySystem to your gem) though cvars is a priceless tool for runtime configuration especially in QA context.


AMZN-alexpete:
Thank you for submitting this new feature @SSPkrolik, it will be very useful!  We've created internal issue LY-90349 to track the progress of getting this into a release.

PR #350, opened 2018-12-10
Enabled EmotionFX Actor to be rendered in camera space

SSPkrolik:
Rendering subsystem of The Engine allows meshes to be rendered in camera space. This PR allows to enable this behaviour for EmotionFX actors.

This feature is especially important for first person games because it allows to not bother with "weapon-in-a-wall" problem which is solved by physics otherwise.

### DIFF
--- a/dev/Gems/EMotionFX/Code/Source/Integration/Assets/ActorAsset.cpp
+++ b/dev/Gems/EMotionFX/Code/Source/Integration/Assets/ActorAsset.cpp
@@ -284,13 +284,24 @@ namespace EMotionFX
 
             pD->m_uniqueObjectId = reinterpret_cast<uintptr_t>(this);
 
+            if (m_isRenderNearest)
+            {
+                rParams.dwFObjFlags |= FOB_NEAREST;
+                pObj->m_ObjFlags |= FOB_NEAREST;
+                rParams.nAfterWater = 1;
+            }
+            else
+            {
+                rParams.dwFObjFlags &= ~FOB_NEAREST;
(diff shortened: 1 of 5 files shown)


dev/Gems/EMotionFX/Code/Source/Integration/Assets/ActorAsset.cpp
```diff
@@ -284,13 +284,24 @@ namespace EMotionFX
...
+                rParams.dwFObjFlags &= ~FOB_NEAREST;
+                pObj->m_ObjFlags &= ~FOB_NEAREST;
+            }
+
             rParams.pMatrix = &m_renderTransform;
 
             pObj->m_II.m_Matrix = *rParams.pMatrix;
             pObj->m_nClipVolumeStencilRef = rParams.nClipVolumeStencilRef;
             pObj->m_nTextureID = rParams.nTextureID;
             rParams.dwFObjFlags |= rParams.dwFObjFlags;
```
alex-leleka: this should be replaced with pObj->m_ObjFlags |= rParams.dwFObjFlags;

AMZN-alexpete:
Hi @SSPkrolik and @alex-leleka, thank you for submitting this feature and feedback.  We've created issue LY-90351 to track this internally and will update the PR when it has been accepted into a release.

CDufour909:
Kinda trivial but you could avoid growing the memory print by 8 bytes (considering class alignment)  in ActorComponent.h by declaring m_isRenderNearest just after m_isRegisteredWithRenderer rather than at the bottom to have the 2 bools packed together.

PR #351, opened 2018-12-11
Enabled rain and snow to be simulated in edit mode

SSPkrolik:
Currently snow and rain are not simulated in Edit mode of the Editor (only in simulation/game mode) which makes very painful to setup snow and rain on level for artists, because for seeing every effect change you have to enter/exit game/simulation mode.


AMZN-alexpete:
Thank you for submitting this feature @SSPkrolik, now we just need to simulate snow plows and umbrellas so artists won't complain about the weather while they work ;)  We've created an internal issue LY-90395 to track the progress of getting this feature in.

PR #352, opened 2018-12-11
Allow to set materials for EMotionFX actor

SSPkrolik:
This allows to use Bus interface for setting Actor materials from code in game.


AMZN-alexpete:
Thank you for submitting this feature @SSPkrolik, we've created internal issue LY-90396 to track it.

PR #353, opened 2018-12-11
Added vertex colors to EMotionFX Actor Builder

SSPkrolik:
Engine's rendering subsystem, shaders and layouts support vertex colors (as well as mesh components) and they are convinient for different rendering tasks.

This PR adds support for vertex colors in EMotionFX Actors.


AMZN-alexpete:
This is an excellent feature @SSPkrolik, thank you!  We've created issue LY-90396 to track it and will update this PR when it is included in a release.

PR #355, opened 2018-12-11
Fixed uninitialized values in actor skinning history data

SSPkrolik:
Fixed uninitialized struct member in actor skinning history data

### DIFF
--- a/dev/Gems/EMotionFX/Code/Source/Integration/Assets/ActorAsset.h
+++ b/dev/Gems/EMotionFX/Code/Source/Integration/Assets/ActorAsset.h
@@ -236,8 +236,8 @@ namespace EMotionFX
             // history for skinning data, needed for motion blur
             struct
             {
-                SSkinningData* pSkinningData;
-                int nFrameID;
+                SSkinningData* pSkinningData{ nullptr };
+                int nFrameID{ 0 };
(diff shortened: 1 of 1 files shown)


dev/Gems/EMotionFX/Code/Source/Integration/Assets/ActorAsset.h
```diff
@@ -236,8 +236,8 @@ namespace EMotionFX
             // history for skinning data, needed for motion blur
             struct
             {
-                SSkinningData* pSkinningData;
-                int nFrameID;
+                SSkinningData* pSkinningData{ nullptr };
+                int nFrameID{ 0 };
```
dreamer-dead: It seems this change will make that unnamed struct to be a non-POD type.
And then it will be an UB to call a memset() at [1]
See for example first part of the [2] topic

[1] https://github.com/aws/lumberyard/blob/2e3184a7d8e59ba05e5707371b8cb6fe40b0ca60/dev/Gems/EMotionFX/Code/Source/Integration/Assets/ActorAsset.cpp#L58
[2] https://stackoverflow.com/a/7189821

SSPkrolik: true, @dreamer-dead . Thanks. will update this.

PR #357, opened 2018-12-12
Added glow (Emittance) map support to particles

SSPkrolik:
Currently engine particle system does not support glow (emittance) map in its setup.
This PR add support for the latter.

Also as shaders are currently bootstrapped and not tracked withing git repo, shader files are added as is with all needed changes.


AMZN-alexpete:
Hi @SSPkrolik, thanks for submitting this feature.  We're tracking this internally with issue LY-90450.  Cheers!

PR #363, opened 2018-12-21
Fixed false alarm warning on level load

SSPkrolik:
This fixes false alarm warning during level load (map command): 

> CEntitySystem::LoadEntities : Not all entities were loaded correctly

which happens when there are no legacy objects in level.pak / mission_0.xml, and it happens when you use only new component entities.


AMZN-alexpete:
Thank you for submitting this fix @SSPkrolik, we've created internal issue LY-90734 to track the progress of getting this into a release.

PR #364, opened 2018-12-26
Fix intersection of water with soft paticle

alex-leleka:
*Problem*
As WaterVolume writes depth, there are happens depth clipping of after water particles that results in hard outlines in point of intersection. 
*Solution*
Enabling soft particle could solve this issue if we update linear depth buffer with water depth.
Moved DepthFixup before water rendering pass. 
Added DepthFixupMerge after water rendering in order to update.
Added linear depth write in water volume color Alpha channel (WaterVolume.cfx).


AMZN-alexpete:
Thank you for submitting this fix @alex-leleka!  We've created an internal issue to track it LY-90756 and will update this pull request when it gets merged into a release.

PR #378, opened 2019-04-03
Fix out of range access for mini table columns

rustamserg:
Number of entries can be invalid in case of columns vector is empty.


AMZN-Hawkeye:
Hi @rustamserg,
Thank you for helping us improve Lumberyard! We are tracking this issue internally as **LY-94779** and we will keep you posted about the status of this PR

PR #379, opened 2019-04-03
Server throttle on linux dedicated server

rustamserg:
Implementation of server throttle on Linux dedicated server.


AMZN-Hawkeye:
Hi @rustamserg,
Thank you for taking the time to help us improve Lumberyard! We will be looking into this Pull Request, and we will keep you updated about the status of the PR. We are tracking it internally as **LY-94781**

PR #380, opened 2019-04-03
Fix mess in profiler output on dedicated server

rustamserg:
Priority 2
profile command output isn't properly rendered on Windows dedicated server console.


AMZN-Hawkeye:
Hi @rustamserg,
Thank you for taking the time to help us improve Lumberyard! We will be looking into this Pull Request, and we will keep you updated about the status of the PR. We are tracking it internally as **LY-94780**

AMZN-alexpete:
Hi @rustamserg , thank you for submitting this improvement.  We've accepted the change in an upcoming version of Lumberyard.

PR #381, opened 2019-04-03
Legacy job executor race condition

rustamserg:
Rare crash caused by race condition in legacy job executor. Trying to fix it by preventing adding new job while job scheduler is in closing state but some jobs are still in progress.


AMZN-Hawkeye:
Hi @rustamserg,
Thank you for taking the time to help us improve Lumberyard! We will be looking into this Pull Request, and we will keep you updated about the status of the PR. We are tracking it internally as **LY-94782**

PR #382, opened 2019-04-03
Script canvas performance optimization

rustamserg:
Priority 1
New entity spawn creates significant performance drop (5-10 fps) on levels with amount of entities is more than 10K. It happens because of copying all loaded entities into new map for Id remapping progress. This fix optimize this routine by copying newly added into already loaded.


AMZN-Hawkeye:
Hi @rustamserg,
Thank you for taking the time to help us improve Lumberyard! We will be looking into this Pull Request, and we will keep you updated about the status of the PR. We are tracking it internally as **LY-94783**

AMZN-puvvadar:
Hi @rustamserg,

This change has been integrated. Thanks for the submission!

PR #407, opened 2019-06-13
V1.18.0.0 voice chat

rustamserg:
Basic in-game voice chat for Windows and Linux platforms. It depends on Microphone Gem and supports default audio device only. Audio is encoded by opus codec and supports 16bits mono stream with 16KHz sample rate.


sourjon:
@rustamserg Thanks, we're taking a look at this!

PR #408, opened 2019-06-21
Bugfix: Freeze when creating slices

CDufour909:
This modification fixes a freeze that can happen when requesting slice instantiation while OnAsset ready is being dispatched.

If the last item in the queue was being Dispatched and this triggered new slice creation requests, then iter was no longer pointing to the end (it was pointing to the old end) and this was causing a freeze while spinning in the loop and progressing the iterator into unknown memory regions.

*Issue #, if available:*

*Description of changes:*


AMZN-Gene:
Awesome find!  How often was this freeze happening?  The slice team will take a look at integrating the fix :)

CDufour909:
Humm... disregard this PR, i found the true root of the issue and will be producing a new fix (and explaining it)

CDufour909:
I created PR 411 with a proper fix for the freeze

bmk10:
how do you force via script for error to happen? I'm also trying to run the editor.exe in profile build in debug mode the whole time. check some my commits for editor.exe crashes I got/get

PR #410, opened 2019-07-03
Fix to Editor resetting the FOV when updating the camera,

Twolewis:
*Issue #, if available:*

*Description of changes:*
Editor updates the camera using default FOV settings, effectively undoing the FOV settings in the camera component.  Fix is to use the existing FOV settings from the existing camera.


AMZN-puvvadar:
Thanks for the change! We'll begin evaluating it as soon as possible.

PR #420, opened 2019-08-05
Fix asset processor batch crash on big number of assets

rustamserg:
Priority 1
Asset processor batch is crashed on big number of assets. The reason of crash is that legacy IDs array is not cleared between assets and constantly growing.


AMZN-puvvadar:
This change has been integrated, thanks for the submission!

PR #421, opened 2019-08-09
Fix uninitialized value in const cvar for release

rustamserg:
*Issue #, if available:*

Some of cvars declared using DeclareConstIntCVar contains uninitialized value in release build. This fix sets such cvars to its default value.
For example cvar: e_MemoryProfiling. For release there is only declaration but initialization is excluded by RELEASE define as result we have random cases when such cvar is not zero and memory profiling info is displayed on the screen


AMZN-Gene:
Nice find!  I'll see about having this pulled into Lumberyard proper.
Thanks

PR #426, opened 2019-09-02
Modify Cry3DEngine to allow terrain alternative via gem(not ready)

caseymcc:
This alters the Cry3DEngine to allow additional terrains systems to be used. I originally performed the modifications on 1.15 and have now ported them to 1.20. I am currently altering it so that the new terrain can be defined in a gem rather that directly (as I did in 1.15). I would like to get some feedback so that this could be integrated back into mainline.


AMZN-nggieber:
Thank you for submitting this pull request.

caseymcc:
Added a few additional fixes for loading/saving properly. Early voxel terrain at https://github.com/caseymcc/Lmbr_VoxelTerrain

PR #427, opened 2019-09-05
Modify waf to build vs solution that actually uses msbuild. 

caseymcc:
Requires setting 'msbuild = True' in user_settings.options under [Visual Studio Project Generator] section


AMZN-nggieber:
Hi @caseymcc, thanks for contributing this to LY we'll take a look at the change.

caseymcc:
For the CryLegacy gem it links to Lua but all of Lua needs to be linked using extern "C" however I see nothing in the files that is in place to handle that. Lua is included with the pre-processor define LUALIB_API (which is also use in CryLegacy) and if it was set to extern "C" would fix the problem however is only seems possible to set it to extern without modifying the code. How is waf handling this?

caseymcc:
In CloudGemFramework why does it matter if the build is a project generator? If it is a generator then is includes lmbraws_unsupported.waf_files no matter if aws_native is supported, even though in lumberyard_sdks.py it list project_generator as supporting aws_native. If lmbraws_unsupported.waf is included in the build then there are lots of things that get defined twice.

PR #430, opened 2019-09-30
Fix MakeDerivedEntity method super class assignment

CVL7:
MakeDerivedEntity always sets its __super reference to BasicEntity, instead of parent entity as it should.


sourjon:
Thanks for the contribution, @CVL7 !  We'll take a look at it!

PR #434, opened 2019-10-11
Fixed CCamera unproject - now always using current camera up.

Twolewis:
CCamera Unproject assumes "up" to be z-up always, which breaks when a non-FPV camera is used.  Fixed to use the current camera's actual up when performing the unproject.


Scott-LY:
Hi @Twolewis, thanks for the pull request. We'll take a look.

PR #440, opened 2019-11-07
Speed up search of assets in Asset Browser

dkondrashkin:
**LY-104815**

*Overal*
- Disabled upward/downward propagation for all filters;
- Added 500 ms delay before **FilterUpdatedSlotImmediate** is called. This allowes to make less filter updates when entering multiple symbols at once;
- Filter invalidation (**invalidateFilter** call) changed to model reset (**begin**/**endResetModel** calls). This is simply cheaper considering the amount of rows in the model;

*In **filterAcceptsRow** method implementation:*
- Added check for the entry's parent being a folder;
- Added cache of previously seen and accepted entries;
- Added recursive children lookup as a replacement for disabled filters propagation option.


sourjon:
Hi @dkondrashkin ,

Thanks for the PR!

PR #444, opened 2019-11-22
Fix of crash in ScriptCanvas Repeater node on Linux Dedicated Server …

sniper-developer:


sourjon:
Thanks @sniper-developer , we appreciate the contribution!  We'll be reviewing it soon!

dev/Gems/ScriptCanvas/Code/Include/ScriptCanvas/Libraries/Core/Repeater.cpp
```diff
@@ -39,7 +39,7 @@ namespace ScriptCanvas
 
                 AZStd::string GetDelaySlotName(Repeater::DelayUnits delayUnits)
                 {
-                    AZStd::string stringifiedUnits = StringifyUnits(delayUnits).c_str();
+                    const char* stringifiedUnits = StringifyUnits(delayUnits).c_str();
```
dreamer-dead: This will not actually fix the crash correctly.
`StringifyUnits()` returns a temporary `AZstd::string` object with a lifetime limited to sequence point at the end of the line.
Storing its content in a `const char*` will not prevent string buffer deallocation.
Later in `AZStd::string::format` deallocated memory will be read, with undefined behaviour.

Fix could look like
```
const AZStd::string& stringifiedUnits = StringifyUnits(delayUnits);
return AZStd::string::format("Delay (%s)", stringifiedUnits.c_str());
```

PR #446, opened 2019-11-25
get_shader_list.py doesn't work when shader list size is bigger than 1MB

rustamserg:
Getting shader list is failed when its size is bigger than 1MB

Uncompressed shader cannot exceed 1MB. This limitation is added as guard into network protocol to ignore shaders which are bigger. But the same network protocol is used for getting shader list which easily can be bigger than 1MB. As result once shader list is grown enough it stops working via get_shader_list.py script.


AMZN-puvvadar:
Thanks for the change @rustamserg , I've forwarded the change to the owning team to consume this request.

PR #447, opened 2019-11-25
Fixes to make shader cache gen works in CI

rustamserg:
Shader cache gen fails on unrelated error, assert spam

After shader list is received the shader cache gen tool tries to release ISystem but not all engine components can be closed gracefully. As result our Jenkins CI plan randomly fails with successfully received shader list. In order to cope we just quit more aggressive via Quit method.
Another issue is related to missed Trace system initialization for shader cache gen application. As result asserts are not handled properly. For example the same assert stack isn't put into ignore list and we get our Jenkins plan overloaded with assert spam.


AMZN-puvvadar:
Thanks for the change @rustamserg , I've forwarded the change to the owning team to consume this request.

PR #448, opened 2019-11-25
LDS memory leaks fix

rustamserg:
Valgrind reports memory leaks on LDS

Fix double allocation in physrea. pArea->m_pMask is already allocated in ProcessBuffer function.
Fix case when client is disconnected while its event queue contains elements.


AMZN-puvvadar:
Thanks for the submission @rustamserg , I will put this in front of our Physics engineers.

PR #449, opened 2019-12-03
Bad render target in snow displacement

yuriy0:
stereoL and stereoR are not resized automatically in
CDeferredShading::CreateDeferredMaps, so it's very likely that stereo{L/R} are
different sizes than the other gbuffer render targets! This breaks things since
all render targets must be the same size.

Also, stereoR is overwritten by SSDO in between snow accumulation and snow
displacement passes. We cannot use stereoR here even if it was the right size.

*Issue #, if available:*

*Description of changes:*
This patch changes the snow displacement render pass to use its own dedicated texture called `$SnowDisplacement` instead of trying to re-use `CTexture::s_ptexStereoR` which doesn't work for several reasons (see above).

It might be the case that there's some other texture which already exists which we can re-use. In this case, this is extremely fragile, because the snow displacement texture is written to right after the z-pass but then used to actually perform snow displacement after deferred tiled lighting, opaque and transparent passes (e.g. water, emissive pass, decals, fog, particles). Any of these draw calls could clobber any "global" texture that we attempt to re-use as the snow displacement RT. So using a dedicated texture is the most robust possible solution.

### DIFF
--- a/dev/Code/CryEngine/RenderDll/XRenderD3D9/D3DDeferredPasses.cpp
+++ b/dev/Code/CryEngine/RenderDll/XRenderD3D9/D3DDeferredPasses.cpp
@@ -673,6 +673,8 @@ bool CD3D9Renderer::FX_DeferredRainGBuffer()
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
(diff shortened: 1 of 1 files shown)


sourjon:
Thanks @yuriy0 !  We appreciate all of your submissions!

PR #450, opened 2019-12-11
Localization init fix to take CVar into account

CDufour909:
Current Localization Initialization defaults to US English even if another language was set to CVar using game.cfg or user.cfg.

I'm rectifying this to check "g_language" CVar and use it if it is set to a relevant language


AMZN-nggieber:
Thank you for submitting this pull request, we will look into it.

PR #454, opened 2019-12-20
Fix for LOD meshes selection in Scene Settings

dkondrashkin:
Broken summary for LODs selection in the newly created elements of Lod Meshes list @ Scene Settings tool.

STR:
1) Start Editor with an empty level;
2) Place entity and add mesh component to it;
3) Assign some asset to the mesh;
4) Go o the Scene Settings (double click on mesh asset);
5) Add a new element to Lod Meshes;
6) Try to select any LOD for the new element.

Result: element summary states: "All Lod meshes selected" no matter what was selected in the last step _except_ for the "Select All/Unselect All" options.

Fix details: 
* Added empty selection list initialization (with unsellect all)


AMZN-nggieber:
@dkondrashkin thank you for submitting this fix! We've integrated the change and it will be available in a future version of Lumberyard.

PR #457, opened 2020-01-08
Added reading vertex normals into render mesh

SSPkrolik:
*Description of changes:*

Added reading vertex normals from geometry description into render mesh. Currently vertex normals are set to defaul (0, 0, 1) vector even if content has vertex normals stream.


AMZN-alexpete:
Hi @SSPkrolik, thank you for submitting this fix/feature.  I've created issue LY-107828 to track this internally and will update this pull request with our progress.

PR #459, opened 2020-01-15
Fixes for AudioEngine gem heap memory corruption on Playstation

simfeo:
Pure new function causes heap corruption for PS4/XboxOne platforms in AudioEngine gem, aznew and azmalloc calls allow us to avoid heap corruption.


AMZN-alexpete:
Hi @simfeo, thank you for submitting this fix!  I've created issue LY-108231 to track this internally and we'll update this pull request with our progress.  Thank you for not submitting code that contains console specific code or code that would break an agreement with a console maker (Sony, Microsoft, Nintendo etc) on this public repository.  If you do want to submit code that does contain such code please contact us through the appropriate support channels.

PR #462, opened 2020-01-29
hash obj folders

SweetId:
.obj files are now stored inside a md5 folder instead of cloning the whole path.
This prevent the engine to fail compiling some files with long path names (or if engine is stored in a long path)


AMZN-nggieber:
@SweetId thank for you submitting this fix, we'll be looking into integrating it.

PR #463, opened 2020-02-03
Animgraph activation performance improvement

KostyaNes:
Animgraph activation performance improvement.


AMZN-alexpete:
Hi @KostyaNes , thanks for submitting this improvement.  We've accepted the change but we will need to write tests after which it will be included in a future version of Lumberyard.  Thanks again!

PR #496, opened 2020-06-26
Replacing pointer comparing to comparing string content

CDufour909:
This fix allows AzPrintF and AzTracePrintf macros to write to log files (Editor.log and game.log) when window is set to "Default"

I think that is was was intended but never worked.

*Issue #, if available:*

*Description of changes:*


AMZN-alexpete:
Thanks for the fix @CDufour909, good catch, we'll take a look!

AMZN-nggieber:
@CDufour909 Thank you, we accepted this change. It should be available in a coming release! We made one change to do a case sensitive strcmp instead of _stricmp.

PR #499, opened 2020-06-29
Bugfix net binding destroying random entities

CDufour909:
When a client is deleting netbound entities (for which he is not the authority), this leaves the NetBinding system with pointers to invalid entities in the m_boundEntities. Causing either crashes or deletion of random entities when the authority actually signals destruction of the netreplicated slice.

To remedy this issue, I'm changingthe m_boundEntities vector contents from Entity pointers to Entity Ids.


AMZN-alexpete:
Thanks for the fix @CDufour909 ! We're looking over the change and will update this pull request as the status changes.

CDufour909:
Hey @AMZN-alexpete,
Been using this code for a couple of days and found a bug where authority deleting a single entity in a slice deleting another entity in the same slice instead. So I just updated the PR to include that fix.

PR #500, opened 2020-07-06
!B Fix stack corruption on non-windows platforms

mik403:
Hi guys!
I am  sharing your fix of stack corruption on non-windows platforms

Problem is next - during asset processing convertor serializes DataType enum as 'int', because this is the underlying type of it on msvc/windows.

But after loading it on Clang  (for example Orbis or Linux) its underlying type will be unsigned int, which will lead to ton's of erros and stack corruption after


AMZN-Gene:
This is great!  Thank you!

PR #501, opened 2020-07-09
Assert fix when setting scrollable entity + avoiding LHS

CDufour909:
Switching scrollable entity leads to asserts in Bus connection code due to the fact that previous handler was not disconnected. So this is a pretty trivial fix.

I also changed the following calls from using the m_scrollableEntity that was just assigned to using the entity is passed as parameter to avoid load-hit-stores.


AMZN-alexpete:
Thanks for submitting this fix @CDufour909! We've integrated the change and it will be part of an upcoming release.

PR #506, opened 2020-07-23
SliceInstantiationTicket hash performance

m9co:
SliceInstantiationTicket is used as bus id in SliceInstantiationResultBus, as result it's used as key for internal hash table and hash function is used extensively when spawning dynamic slices.
Current implementation of hash function uses Crc32, which is quite slow - on my PC it takes around 20 microseconds, which can add up to a considerable time loss - single slice spawn causes multiple hash table operations, recalculating hash each time.


AMZN-alexpete:
Thanks for submitting this performance boost @m9co !  We're currently evaluating this change.

AMZN-alexpete:
Hi @m9co , this change has been accepted and will be released in a future version of Lumberyard, thank you!

PR #494, opened 2020-06-04
Statistics collection and thread safety fixes for AZ::IO::DedicatedCache

dkondrashkin:
This commit addresses 2 problems we've encountered:
1. Broken RAD telemetry plots (collected statistics) for AZ::IO::Streamer;
2. Rare crashes of game client.

For the first issue we have a following visualization:
image

Streamer stat names appeared to be corrupted. This happened because of Statistic::CreateXXX(...) functions, accepting AZStd::string_view as a first argument, were supplied with AZStd::string variables living in method scope. To fix this we've introduced a separate cache for stat names.

Second issue (crash) was preceded by the following assertion:

`<2020-05-29T18:22:20:529+03> (System) - Trace::Assert ...\Code\Framework\AzCore\AzCore/std/containers/vector.h(585): (14664) 'const class std::unique_ptr<class AZ::IO::BlockCache,struct std::default_delete<class AZ::IO::BlockCache> > &__cdecl AZStd::vector<class std::unique_ptr<class AZ::IO::BlockCache,struct std::default_delete<class AZ::IO::BlockCache> >,class AZStd::allocator>::operator [](unsigned __int64) const'`
`<2020-05-29T18:22:20:529+03> (System) - AZStd::vector<>::at - position is out of range`
`<2020-05-29T18:22:20:530+03> (System) - ------------------------------------------------`
`<2020-05-29T18:23:37:320+03> (System) - 00007FF7B9793A27 (game01Launcher) : AZ::IO::FullFileDecompressor::CollectStatistics`
`<2020-05-29T18:23:37:321+03> (System) - 00007FFC2188AEE9 (CryRenderD3D11) : AZ::IO::Device::CollectStatistics`
`<2020-05-29T18:23:37:321+03> (System) - 00007FFC21890D54 (CryRenderD3D11) : AZ::IO::Device::OnTick`
`<2020-05-29T18:23:37:321+03> (System) - 00007FF7B968B787 (game01Launcher) : AZ::Internal::EBusContainer<AZ::TickEvents,AZ::TickEvents,0,2>::Dispatcher<AZ::EBus<AZ::TickEvents,AZ::TickEvents> >::Broadcast<void (__cdecl AZ::TickEvents::*)(float,AZ::ScriptTimePoint) __ptr64,float & __ptr64,AZ::Sc`
`<2020-05-29T18:23:37:321+03> (System) - 00007FF7B96E57C7 (game01Launcher) : AZ::ComponentApplication::Tick`
[...]

PR #507, opened 2020-08-05
Statistics collection and thread safety fixes for AZ::IO::DedicatedCache

dkondrashkin:
More fixes for problems addressed by https://github.com/aws/lumberyard/pull/494

While previous pull request fixed most of the issues the one remained:
* `DedicatedCache::DestroyDedicatedCache()` call can happen right between `DedicatedCache::CollectStatistics()` call and any code that makes use of statistics collected (e.g. `System.cpp @ 1745` or `Device.cpp @ 96`). This leads to accessing invalid memory through string views stored in statistics values and, particularly, to crashes on consoles.

The solution proposed (apart form guarding internal data members with locks) is to postpone removal of `m_cachedStatNames` entries until the very beginning of the next `DedicatedCache::CollectStatistics()` call.


AMZN-alexpete:
Thanks for submitting these additional fixes @dkondrashkin, we're looking over the change and will update this PR with status changes or questions.

PR #508, opened 2020-08-10
adding vs2019 bin folders in the  .gitignore file

ravenboilinux:
*Description of changes:*
I noticed that in the .gitingore file that the Lumberyard Team added that bin folders for VS2017 but not for VS2019 so this PR just added that folder in right under the folders for VS2017

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -105,6 +105,9 @@
 /dev/Bin64
 /dev/Bin64vc141
 /dev/Bin64vc141.Debug
+/dev/Bin64vc142
+/dev/Bin64vc142.Debug.Dedicated
+/dev/Cache
 /dev/Cache
 /dev/SamplesProject
 /dev/Code/SDKs
@@ -130,3 +133,4 @@
 /dev/SetupAssistantUserPreferences.ini
 /dev/_WAF_/environment.json
 /SetupAssistant.bat
+


AMZN-Gene:
Nice find!  Thank you 👍

.gitignore
```diff
@@ -105,6 +105,9 @@
 /dev/Bin64
 /dev/Bin64vc141
 /dev/Bin64vc141.Debug
+/dev/Bin64vc142
+/dev/Bin64vc142.Debug.Dedicated
+/dev/Cache
```
AMZN-alexpete, line 110: Looks like this got duplicated

.gitignore
```diff
@@ -105,6 +105,9 @@
 /dev/Bin64
 /dev/Bin64vc141
 /dev/Bin64vc141.Debug
+/dev/Bin64vc142
+/dev/Bin64vc142.Debug.Dedicated
```
AMZN-alexpete, line 109: replacing all /dev/BinXXXXX entries with a single /dev/Bin* would be simpler and handle for future versions.

AMZN-alexpete:
Hi @ravenboilinux, thanks again for sending this in.  As it turns out we have this change internally, however there was an issue in our pipeline that prevented it from propagating to GitHub.  We're working on fixing that now.  Cheers!

PR #512, opened 2020-08-28
Add unbounded overlap to physics world

yuriy0:
This change exposes the 'unbounded' mode for PhysX overlap queries.

This is much more efficient that using the regular (bounded) overlap if: you're querying very large areas which have very many physical objects inside, and you aggregate the result of the query immediately as opposed to operating on individual objects returned from the query.

*Issue #, if available:*

*Description of changes:*


AMZN-puvvadar:
Hi @yuriy0 ,

Thanks for the submission, we'll evaluate it and update when we have more info.

AMZN-puvvadar:
Hi @yuriy0 ,

We've evaluated the change and will be accepting pending some performance testing. Thanks for the submission!

PR #513, opened 2020-08-28
Add the ability to override how motion extraction is handled

yuriy0:
Motion extraction from EMFX is currently handled one of three ways:
- PhysX character controller
- Legacy CryPhysics character controller
- TransformBus

Sometimes none of these methods are desirable. It is useful to be able to customize the handling of motion extraction depending on game mechanics. For example we might want to apply a maximum velocity to motion extracted movement.

*Issue #, if available:*

*Description of changes:*
This change adds a new bus for handling motion extraction, and adds a new check for the presence of a handler for this bus in the motion extraction logic.


AMZN-puvvadar:
Hi @yuriy0 ,

We've accepted this change and will be planning to integrate it (with some minor changes). Thanks for the submission!

PR #515, opened 2020-08-28
Only update PhysX editor colliders if transform has actually changed

yuriy0:
In the editor, PhysX colliders cannot be moved and are instead destroyed and re-created in the physical world every time the transform changes. This update is a relatively expensive operation which adds up quickly if the transform is updated very frequently e.g. every tick, and you have many colliders in your level.

This change adds a check in OnTransformChanged to see if the transform is actually different.

*Issue #, if available:*

*Description of changes:*


AMZN-puvvadar:
Hi @yuriy0 ,

Thanks for submitting this optimization, we'll evaluate it and update when complete.

oyongzhi:
Hello yuriy0! We've reviewed your pull request and wanted to let you know we will be incorporating your proposed changes into our next version of the Lumberyard engine. Thank you kindly for contributing to Lumberyard!

PR #516, opened 2020-08-28
Don't serialize PhysX collider scale values

yuriy0:
The scale values for PhysX colliders come from the entity parent transform, so when changing the scale anywhere in the parent hierarchy of a physx collider, the editor thinks thinks that the collider has changed, because this changes its serialized representation, and comparison is done based on the serialized representation.

It isn't necessary to store the scale in the serialized representation for runtime correctness, because all of the runtime colliders invariably overwrite the scale of their shapes. This is necessary for the correct instantiation of dynamic slices which might a scale override.

However instead of removing the scale from the serialized representation altogether this change simply sets the scale only when creating the collider in the PhysX world, not in the serialized representation. The serialized representation then always contains (1,1,1) as the scale.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/dev/Gems/PhysX/Code/Source/EditorColliderComponent.cpp
+++ b/dev/Gems/PhysX/Code/Source/EditorColliderComponent.cpp
@@ -471,6 +471,9 @@ namespace PhysX
 
             m_editorBody = AZ::Interface<Physics::System>::Get()->CreateStaticRigidBody(configuration);
 
+            auto& shapeConfiguration = m_shapeConfiguration.GetCurrent();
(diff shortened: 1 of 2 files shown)


AMZN-puvvadar:
Hi @yuriy0 ,

Thanks for the submission. We'll update once we've evaluated this change.

AMZN-puvvadar:
Hi @yuriy0,

We acknowledge the issue this change is addressing but will be going a different route to fix it. Thanks for bringing this to our attention!

PR #517, opened 2020-08-28
Correctly upgrade PhysX mesh asset description

yuriy0:
Version 1 had an 'export as convex' field which maps to 'export method' in Version 2. The existing version converter does not handle the 'export as convex' field, which causes physx colliders being upgraded from Version 1 to Version 2 to always have the default 'export method'.

*Issue #, if available:*

*Description of changes:*


oyongzhi:
Hello yuriy0! We've reviewed your pull request and wanted to let you know we will be incorporating your proposed changes into our next version of the Lumberyard engine. Thank you kindly for contributing to Lumberyard!

PR #519, opened 2020-08-28
Allow configuration of Actor BBox update type in editor

yuriy0:
ActorInstance has several different modes and configuration values for computing its bounding box. This is especially useful when some animations might move the actor far outside its static bounds, in which case the actor render node will dissapear even when the actor mesh is still visible.

These settings are not exposed to the frontend (ActorComponent). This change simply adds these settings to the Actor configuration and to the Editor Actor edit context.

*Issue #, if available:*

*Description of changes:*


AMZN-puvvadar:
Hi @yuriy0 ,

Thanks for the submission! We'll review it internally and get back to you.

AMZN-puvvadar:
Hi @yuriy0 ,

We're planning to integrate this fix. Thanks for the submission!

PR #522, opened 2020-09-28
Wrong PhysX scene query buffer size

m9co:
When performing scene query with more requested hits than maximum buffer size, passed buffer size (request.m_maxResults) will be lower than actual buffer size and may cause crashes.

### DIFF
--- a/dev/Gems/PhysX/Code/Source/World.cpp
+++ b/dev/Gems/PhysX/Code/Source/World.cpp
@@ -314,7 +314,7 @@ namespace PhysX
             s_raycastBuffer.resize(maxResults);
         }
         // Raycast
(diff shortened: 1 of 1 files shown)


AMZN-alexpete:
Thanks for submitting this fix @m9co, we're looking into integrating this.

SergeyAMZN:
Hello @m9co ! We've reviewed your pull request and wanted to let you know we will be incorporating your proposed changes into our next version of the Lumberyard engine. Thank you kindly for contributing to Lumberyard!

wcbx:
You should tag and close PRs you integrate.

m9co:
@wcbx should I be doing that with my PRs?

wcbx:
@m9co you can do what you want with your branch. I was saying Amazon should close this if they integrated it or at least tag it as integrated. Ideally they would do both.

PR #523, opened 2020-09-30
Fix compiling error in The Asian version of VS

GameInstitute:
*Issue #, if available:*

*Description of changes:*


AMZN-alexpete:
Thank you for submitting this fix @GameInstitute, we're looking at integrating this fix.

PR #533, opened 2020-12-20
EMotionFX crash when used with replication

SecretsOTheP:
*Issue #, if available:* N/A

*Description of changes:* Fixes an uninitialized variable in an EmotionFX component. (m_animGraphInstance)

This variable will cause a crash, especially in a replicated situation where Activate is not called on this component - as it doesn't use a weak pointer container. 

m_animGraphInstance needs to be set to null before being used - otherwise, the pointer will be considered valid memory as it's set to a non-null, garbage memory value by default.


AMZN-Gene:
Thanks for the fix! I’ll pass this to the character team for review :)

PR #535, opened 2021-01-11
V1.27.0 emotionfx memory fixes

mik403:
These are set of memory improvements, as discussed on the private forum, LY-23014
See discussion topic for more info and results.

Next changes were made:

- Decreased translation precision on MotionDataBuilder
- Shrink shared data in Importer to reclaim some memory
- Change reservation strategy in MotionInstancePool to reduce fragmentation
- Released unused raw asset data in EmotionFX asset to reclaim memory after asset initialization
- Removed GPU deformed meshes from CPU after its skinning


AMZN-Gene:
Thanks, this is great! I'll work on getting these optimizations into LY Vanilla!

dev/Gems/EMotionFX/Code/EMotionFX/Source/Importer/Importer.cpp
```diff
@@ -634,6 +634,7 @@ namespace EMotionFX
         // get rid of shared data
         ResetSharedData(sharedData);
         sharedData.Clear();
+        sharedData.Shrink();
```
AMZN-Gene, line 637: Is the clear/shrink needed since sharedData is going out of scope and will be destroyed?

PR #539, opened 2021-01-26
V1.27 emfx param motion node

DDR4eG:
This PR adds a new anim graph state machine node, which takes motion name as an input string parameter, searches it in a current motion set, and plays motion similar to AnimGraphMotionNode.

image


AMZN-Gene:
Very cool! Is there an example use case you'd recommend for using this type of node? What happens if there aren't any motions in the set matching?

DDR4eG:
Hi @AMZN-Gene,
I've left a more detailed comment on this with an example in the original forum post (ticket #LY-112755).

An empty motion will be played (T-pose for characters) if there is no motion with a given name in current motionset.

PR #540, opened 2021-01-26
EmotionFX: Ability for SimpleMotionComponent & AnimGraphComponent to exist on same entity

DDR4eG:
This PR makes it possible to use both SimpleMotion and AnimGraph components on the same entity.
Such actor setup can be used for animating character emotes, in cutscenes, etc.


AMZN-Gene:
Nice! I could also see this being useful for debugging. Sometime I want to quickly make sure a particular animation looks good, but have to either open up emfx editor and play the anim, or delete the animgraph component, add the simple component, play the simple component, and then swap them back again. Thanks!

PR #541, opened 2021-02-08
Add auth header and upgrade SSL version to sslv23

rustamserg:
*Issue:*
It is not possible to provide Authorization header to setup secure access into websocket gateway exposed to public internet.
Generic TLS version 1 is not accepted by modern nginx setup so changed it into sslv23.

*Description of changes:*
These changes are made to keep diff as small as possible to show places where gem's clients need modifications. It is not a good idea to take it as is but it is good starting point to understand what should be exposed via public gem bus.
So it would be awesome to have possibilities to add own headers and configure SSL/TLS versions for secure connections.


AMZN-nggieber:
@rustamserg Thank you for the PR, well look into your starting point and investigate implementing some of your suggestions!

